PR #85, opened 2018-08-08
9 measuring tool

neodyymi:
#9 rest of the issue


neodyymi:
Doesn't work correctly on Chrome. Works on Edge and Firefox.

neodyymi:
Seems to work fine on Chrome now.. Not sure why it didn't before and why it does now.

PR #125, opened 2018-08-22
115 toolbar

kronken:


src/components/map/Toolbar.tsx
```diff
@@ -0,0 +1,75 @@
+import * as s from './toolbar.scss';
+import React, { Component } from 'react';
+import { FiEdit, FiCopy, FiPlusSquare, FiPrinter } from 'react-icons/fi';
+import ToolbarIcon from './ToolbarIcon';
+
+interface ToolbarState {
+    editModeIsActive: boolean;
```
Divinare: isEditModeActive? (We have isMapFullScreen)

I think better might be "isRouteEditable" but not sure yet because we aren't that far yet. So no need to change.

src/components/map/Toolbar.tsx
```diff
@@ -0,0 +1,75 @@
...
+
+        const addNode = () => {
+        };
+
+        const print = () => {
+        };
+
+        return (
+            <div className={s.toolbar}>
+                <ToolbarIcon
```
Divinare: Rename ToolbarIcon.tsx -> ToolbarButton.tsx?

src/components/map/Toolbar.tsx
```diff
@@ -0,0 +1,75 @@
...
+                    onClick={editRoutes}
+                    isActive={this.state.editModeIsActive}
+                    isDisabled={false}
```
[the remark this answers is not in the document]
kronken: This is a nice idea, I was initially thinking to have different handlers for the different buttons. But having a isActive = {selectedTool.name === 'stuff'} thingy is smarter.

Issue #91, opened 2018-08-14
RouteStart markers


Divinare:
Draw on map: Marker + number to the beginning of each routePath

https://app.zeplin.io/project/5b054c19f46b8229769ca6cb/screen/5b054c4a7aaa417d22292860

PR #126, opened 2018-08-22
91 routestart markers

ahjyrkia:
https://github.com/HSLdevcom/jore-map-ui/issues/91


src/layers/nodeMarker.ts
```diff
@@ -14,7 +16,16 @@ export default class NodeMarker {
...
         this.options = options;
 
-        const icon = this.getIcon(this.options.color);
+        let icon;
+        if (this.options.type === NodeType.START) {
+            const divIconOptions : L.DivIconOptions = {
+                className: s.nodePin,
+                html: `<div class="${s.pin}" style="--pinColor: ${this.options.color};"/>`,
+            };
+            icon = new L.DivIcon(divIconOptions);
```
Divinare: Instead of L.DivIcon + making marker with pure CSS, using .svg would be better:
https://fontawesome.com/icons/map-marker-alt?style=solid

L.Marker takes an icon as parameter:
https://leafletjs.com/reference-1.3.4.html#marker-option
https://leafletjs.com/reference-1.3.4.html#icon

src/factories/nodeFactory.ts
```diff
@@ -15,6 +15,18 @@ class NodeFactory {
...
         };
     }
+
+    public static createStartingPointNode = (coordinates: any): INode => {
+        const coordinate : ICoordinate = {
+            lon: coordinates[0],
+            lat: coordinates[1],
+        };
+        return <INode>{
+            id: 0,
```
Divinare: Not sure if hard coded id is good. Maybe use incremental number as id?

src/factories/routePathFactory.ts
```diff
@@ -2,17 +2,24 @@ import { INode, IRoutePath } from '../models';
...
     public static createRoutePath = (suunta: any, isVisible:boolean): IRoutePath => {
         const nodes:INode[]
         = suunta.reitinlinkkisByReitunnusAndSuuvoimastAndSuusuunta.edges
-            .map((node: any) => NodeFactory.createNode(node.node));
+            .map((node: any) => {
+                return NodeFactory.createNode(node.node);
+            });
+
+        const geoJson = JSON.parse(suunta.geojson);
+        const asd = NodeFactory.createStartingPointNode(geoJson.coordinates[0]);
```
Divinare: nodes.push(NodeFactory.createStartingPointNode(geoJson.coordinates[0]);?

The order of nodes shouldn't matter when they have NodeType that separates them. Just prefer using push over unshift

Divinare:
There will be a lot of changes coming from PR https://github.com/HSLdevcom/jore-map-ui/pull/135 

Instead of merging, consider of making these changes again to a fresh branch from master

PR #130, opened 2018-08-27
116 map layers button

Divinare:
https://github.com/HSLdevcom/jore-map-ui/projects/1

Tried to make hover with just CSS. Doesn't quite work. If we don't get it to work, revert to that commit 
a40fb00?


Divinare:
You can review this now.

src/components/map/MapLayersControl.tsx
```diff
@@ -0,0 +1,56 @@
...
+    selectedOption: string;
+}
+
+@observer
+export default class MapLayersControl extends React.Component
+<INodeWindowProps, IMapLayersControlState> {
+    constructor (props: any) {
+        super(props);
+        this.state = {
+            selectedOption: 'kartta',
```
kronken: I would like this to be an enum instead of hard-coded strings.
No need to put the enum in the enum folder, the enum could well be specified locally in this file since it is only used locally.

Divinare:
This can be reviewed again.

PR #133, opened 2018-08-28
63 react router and loading animation

### LINKED ISSUE #63: Sidebar store: state handling
sidebar store:

isLoading<boolean>: true/false
errors<String[]>
view<String>: "lineList / lineEdit"

neodyymi:
closes #63, closes #131 and closes #132

- Working routing and loading state.
- Mobx-errors fixed. (this was a weird one..)
- Fixed link colors
- route parsing with qs-library

Issues left from #132:
- make all getRoute calls in one grapQL query (now it makes separate calls for each routes)
- removing route


src/components/App.tsx
```diff
@@ -14,26 +13,64 @@ import Map from './map/Map';
...
+        } catch (err) {
+            // TODO: show error on screen that the query failed
+        }
+    }
+    private async queryRoutes() {
+        this.props.lineStore!.setSearchInput('');
+        const queryValues = qs.parse(this.props.location.search,
+                                     { ignoreQueryPrefix: true, arrayLimit: 1 },
+        );
+        let routeIds: string[] = [];
```
Divinare: nit: put let routeIds inside if

src/components/sidebar/LineItem.tsx
```diff
@@ -1,40 +1,41 @@
...
-                this.props.lineStore!.setSearchInput('');
-            })
-            .catch((err: any) => {
-            });
-    }
-
     public render(): any {
+        let pathname = (this.props.location.pathname === '/')
+            ? 'routes/' : this.props.location.pathname;
+        pathname += '';
```
Divinare: Why add an empty string to pathname?

src/components/sidebar/LineItems.tsx
```diff
@@ -9,23 +9,16 @@ import * as s from './lineItems.scss';
...
-                })
-                .catch((err: any) => {
-                    // tslint:disable-next-line:no-console
-                    console.log(err);
-                });
-        }
+    async componentWillMount() {
+        this.props.lineStore!.linesLoading = true;
+        await this.props.lineStore!.setAllLines(await LineService.getAllLines());
+        this.props.lineStore!.linesLoading = false;
```
Divinare: Why we have a line service call in LineItems?

And why lineStore has linesLoading? Isn't one loading variable enough in sidebarStore?

src/components/sidebar/LineSearch.tsx
```diff
@@ -22,6 +19,7 @@ class LineSearch extends React.Component<ILineSearchProps, ILineSearchState> {
         this.state = {
             lineItems: '',
         };
+        this.props.lineStore!.setSearchInput('');
```
Divinare: Is this necessary?

src/components/sidebar/Sidebar.tsx
```diff
@@ -1,50 +1,46 @@
...
-                    {this.props.routeStore!.routes.length < 1 &&
-                        <LineSearch
-                            showSearchResults={true}
-                        />
-                    }
-                    {this.props.routeStore!.routes.length > 0 &&
-                        <RoutesEdit />
-                    }
-                </div>
+                <Route component={SidebarContent} />
```
Divinare: Why split current Sidebar into Sidebar & SidebarContent?

src/components/sidebar/Sidebar.tsx
```diff
@@ -1,50 +1,46 @@
...
 
-    public handleHeaderClick = () => {
-        this.props.routeStore!.clearRoutes();
-    }
-
     public render(): any {
+        const handleHeaderClick = () => {
+            this.props.routeStore!.clearRoutes();
+            this.props.lineStore!.setSearchInput('');
+            this.props.lineStore!.lineSearchVisible = true;
```
Divinare: LineSearch is always visible, remove this line

src/stores/lineStore.ts
```diff
@@ -5,10 +5,14 @@ export class LineStore {
     @observable private _filters: string[];
     @observable private _allLines: ILine[];
     @observable private _searchInput: string;
+    @observable private _linesLoading: boolean;
+    @observable private _lineSearchVisible: boolean;
```
Divinare: Remove this these variables from store?

src/components/sidebar/lineSearch.scss
```diff
@@ -1,9 +1,5 @@
 .lineSearchView {
-    max-height: 100%;
-    width: 100%;
-    display: flex;
-    flex-direction: column;
-
+
```
Divinare: Remove styles from here that are used in searchResults.scss

src/components/sidebar/SidebarContent.tsx
```diff
@@ -0,0 +1,48 @@
...
+        if (this.props.sidebarStore!.isLoading) {
+            return (
+                <div className={s.sidebarContentView}>
+                    <div id={s.loader} />
+                </div>
+            );
+        }
+        return (
+            <div className={s.sidebarContentView}>
+                <LineSearch/>
```
Divinare: How about this:
if (this.props.sidebarStore!.isLoading) {
    <Loader />
}

src/components/App.tsx
```diff
@@ -14,26 +13,64 @@ import Map from './map/Map';
...
+    componentWillReceiveProps(props: any) {
+        this.queryAll();
+    }
+
+    async queryAll() {
+        this.props.sidebarStore!.isLoading = true;
+        await this.queryAllLines();
+        await this.queryRoutes();
+        this.props.sidebarStore!.isLoading = false;
+    }
```
kronken: Doesn't this mean that we are querying all lines and routes, every time the url changes? I know that Apollo doesn't actually send the request do to caching, but still? Overall I don't think that the App component should take care of fetching things that some of it children might need.

src/components/sidebar/LineItems.tsx
```diff
@@ -9,23 +9,16 @@ import * as s from './lineItems.scss';
 
 interface ILineItemsProps {
     lineStore?: LineStore;
-    searchInput: string;
+    location: any;
```
kronken: We should not push location prop down all the way through the tree of components. Overall, I don't think LineItem should need to worry about location stuff at all.

src/components/sidebar/LineItem.tsx
neodyymi: Removed

src/components/sidebar/LineItems.tsx
neodyymi: Removed as old and unnecessary

src/components/sidebar/LineSearch.tsx
neodyymi: Apparently not anymore :)

src/components/sidebar/Sidebar.tsx
neodyymi: True dat

src/stores/lineStore.ts
neodyymi: Yup, removed

src/components/App.tsx
neodyymi: I do somewhat agree.. Let's discuss this at the office

src/components/sidebar/LineItems.tsx
neodyymi: I've used it in somecases as a workaround for issues with mobx and react-router. It's helped in cases where site isn't getting updated when url changes.

neodyymi: That said, I agree on having less information pushed to the furthest components and maybe this can be resolved.

neodyymi: In this case it's for the LineItem to be able to form urls for displaying chosen routes.

kronken:
Before merge:
- Use loading component
- LineSearch -> mounted: getAllRoutes()
- RouteList -> mounted: getRoutes()
- RouteParsing: use urlService
    - Just move buildPath 
- Make cards to our backlog for after merge tasks

After merge:
- Route parsing
- make all getRoute calls in one grapQL query (now it makes separate calls for each routes)
- removing route

Issue #113, opened 2018-08-21
RoutePaths for search result


Divinare:
- Clicking down arrow icon showns routePath list
- Each routePath has:
   - A checkbox (use our generic checkbox component)
   - 2 timestamps:
          Voim.ast 20.8.2018
          Viim.voim.olo 29.8.2018

PR #134, opened 2018-08-29
113 routepaths in search

ahjyrkia:
https://github.com/HSLdevcom/jore-map-ui/issues/113


kronken:
Overall it works well, however, I found the LineItem.tsx file to be a bit too large, and a bit too difficult to read. I did a small refactor of this PR, where I splitted the LineItem.tsx, and set the RoutePaths informations into a new file called LineItemSubMenu.tsx. I also moved the select logic into this file, which allow us to have multiple routePaths open at once. 

I didn't want to commit straight to this branch, since it changes rather much. However, I created a PR for merging my changes into this 113-routepaths-in-search branch. My changes can be found in: https://github.com/HSLdevcom/jore-map-ui/pull/136 . Feel free to merge these changes if you feel they are good.

Divinare:
I would do some minor UI improvements:
- Arrow should direct right when not opened, arrow should direct down when opened
- "Muokattu" text starts from different place than that blue clickable link (2px difference)
- add padding-top: 5px to routeDate
- Check background-color of routeName when hovered. That padding should be fixed (its because now routeItem has 10px margin left)

src/components/sidebar/LineItemSubMenu.tsx
```diff
@@ -0,0 +1,123 @@
...
+            this.unSelect(index);
+        } else {
+            this.select(index);
+        }
+    }
+
+    render () {
+        if (this.props.visible && this.state.routePaths == null) {
+            this.fetchRoutePaths();
+        }
```
Divinare: This check belongs to componentWillReceiveProps I suppose?

src/components/sidebar/LineItem.tsx
```diff
@@ -1,39 +1,122 @@
...
+    private toggleRouteMenu(routeId: string, e: any) {
+        e.stopPropagation();
+        if (this.isRouteOpen(routeId)) {
+            this.closeRouteMenu(routeId);
+        } else {
+            this.openRouteMenu(routeId);
+        }
+    }
+
+    private getRoute(route: ILineRoute) {
```
Divinare: How about renderRoute? Now getRoute() gives the idea of fetching route from GraphQL.

For methods that return JSX, i would put render prefix

src/components/sidebar/LineItemSubMenu.tsx
```diff
@@ -0,0 +1,123 @@
...
+
+        if (!this.props.visible) {
+            return null;
+        }
```
Divinare: Is this necessary? We already have:

interface lineItemSubMenuProps {
    routeId: string;
    lineId: string;
    visible: boolean;
    notificationStore?: NotificationStore;
}

That "?" after notificationStore means it can be null but without "?" typescript should give error if the prop is not given?

Issue #121, opened 2018-08-21
Only allow dragging when in edit-mode


Divinare:
- Pen icon should go active when clicked

PR #135, opened 2018-08-29
121: Routes and nodes as react components & allow dragging

kronken:
#121


src/factories/nodeFactory.ts
```diff
@@ -2,14 +2,15 @@ import { INode, ICoordinate } from '../models';
...
-    public static createNode = (node: any): INode => {
+    public static createNode = (routePathId: string, node: any): INode => {
         const coordinateList = JSON.parse(node.solmuByLnkalkusolmu.geojson);
         const coordinate : ICoordinate = {
             lon: coordinateList.coordinates[0],
             lat: coordinateList.coordinates[1],
         };
 
         return <INode>{
+            routePathId,
```
Divinare: Should this be renamed so that its clear that its generated id and not from jore?

I would rename as:
internalRoutePathId
or
generatedRoutePathId

src/factories/routePathFactory.ts
```diff
@@ -1,18 +1,39 @@
...
 
 class RoutePathFactory {
     // suunta to IRoutePath
-    public static createRoutePath = (suunta: any, isVisible:boolean): IRoutePath => {
+    public static createRoutePath = (
+        routeId: string,
+        suunta: any,
+        isVisible:boolean,
+    ): IRoutePath => {
+        const madeUpId = HashHelper.getHashFromString(
```
Divinare: This should be renamed as the same variable as used in NodeFactory's createNode

src/factories/routePathFactory.ts
```diff
@@ -1,18 +1,39 @@
...
+            .map((node: any) => NodeFactory.createNode(madeUpId, node.node));
+
+        const coordinates = JSON.parse(suunta.geojson).coordinates;
+        const positions = coordinates.map((coor: [number, number]) => [coor[1], coor[0]]);
 
         return <IRoutePath>{
+            routeId,
             nodes,
+            positions,
+            internalRoutePathId: madeUpId,
```
Divinare: Here it is as "internalRoutePathId" but in NodeFactory its "routePathId". We need consistency, use the same name everywhere.

kronken:
Fixes from review done, can be reviewed again

PR #150, opened 2018-09-04
Use react style refs to be able to use componentDidUpdate instead of …

kronken:
…componentWillReceiveProps


Divinare:
Lets try with getMap() returning leafletElement

src/components/map/RouteLayer.tsx
```diff
@@ -10,11 +11,11 @@ interface RouteLayerProps {
 }
 
 export default class RouteLayer extends Component<RouteLayerProps> {
-    calculateBounds(nextProps: RouteLayerProps) {
+    calculateBounds() {
```
Divinare: Nit: always use private/public? Use here private right?

Issue #114, opened 2018-08-21
Search results: selecting routePaths


Divinare:
- Add button to the end of search component
   - This button should be visible all the time
   - Button has either "Lisää (3)" or "Sulje" text
         - Lisää button is green/blue, sulje button is grey
- Clicking "Lisää" adds only selected routePaths to showRoute view.
   - Add routes to routeStore that have only the selected routePaths

- Clicking "Sulje" closes search. Clear search input field also

PR #151, opened 2018-09-04
114 selecting route paths from home view

Divinare:
Resolves: https://github.com/HSLdevcom/jore-map-ui/issues/114


src/components/sidebar/LineItemSubMenu.tsx
```diff
@@ -4,42 +4,52 @@ import { IRoutePath, IRoute } from '../../models';
...
         this.state = {
             routePaths: null,
-            checkedRoutePaths: [],
+            selectedIds: [],
         };
     }
 
+    componentWillUnmount() {
+        this.mounted = false;
+    }
```
kronken: Why do we need this.mounted?

src/components/sidebar/LineItemSubMenu.tsx
```diff
@@ -49,31 +59,36 @@ class LineItemSubMenu extends Component<LineItemSubMenuProps, LineItemSubMenuSta
...
 
-    private select(index: number) {
+    private select(routePathId: string) {
+        const newSelectedIds = this.state.selectedIds;
+        newSelectedIds.push(routePathId);
         this.setState({
-            checkedRoutePaths: this.state.checkedRoutePaths.concat(index),
+            selectedIds: newSelectedIds,
         });
+        this.props.searchStore!.addSubLineItem(this.props.routeId, routePathId);
```
kronken: We should probably avoid using both store and local state, when the data we handle basically is the same. What if the data only would be in the store, and a routePath could check directly from there if it is checked or not?

Divinare:
Fixes done. This can be re-reviewed.

PR #154, opened 2018-09-04
additions to tslint

### LINKED ISSUE #144: tslint-rule for order of imports
Also fix project to pass linter

neodyymi:
closes #144


Divinare:
The diff should show only changes made into this PR. Is there a way to fix?

Also are those new rules? Comment here or to the tslint file?

Divinare:
// test how this works (use componentWillMount or..)
    "deprecation": true,

    // order: static variables, variables, constructor, lifecycle, public/private, static methods
    "member-ordering": [true, {
      "order": "fields-first"
    }],

PR #168, opened 2018-09-04
91 routestart markers v2

ahjyrkia:
https://github.com/HSLdevcom/jore-map-ui/issues/91


src/components/map/NodeLayer.tsx
```diff
@@ -52,23 +55,52 @@ export default class NodeLayer extends Component<NodeLayerProps> {
...
-            const icon = this.getIcon(node);
+        const colors = ColorScale.getColors(this.props.firstNodes.length);
+        return (
+            <div>
+            {
+                this.props.firstNodes.map((coordinates, index) => {
+                    const icon = this.getStartPointIcon(colors[index]);
+                    return (
+                        <Marker
+                            draggable={true}
```
kronken: Dragging should probably be set to false.

src/components/map/Map.tsx
```diff
@@ -104,6 +104,18 @@ class LeafletMap extends React.Component<IMapProps, IMapState> {
         );
     }
 
+    private getFirstNodes() {
```
kronken: We could simplify this since this function is mostly getting the visible route paths, but we already have this information in the visibleRoutePaths variable. So, what if we instead would pass these visibleRoutePaths to this getFirstNodes function, and it would just map these visibleRoutePaths by returning their start locations. Something like this:

    private getFirstNodes(visibleRoutePaths: IRoutePath[]) {
        return visibleRoutePaths.map(rPaths => rPaths.geoJson.coordinates[0]);
    }

src/components/map/Map.tsx
```diff
@@ -149,6 +160,7 @@ class LeafletMap extends React.Component<IMapProps, IMapState> {
                     />
                     <NodeLayer
                         nodes={visibleNodes}
+                        firstNodes={this.getFirstNodes()}
```
kronken: Would a more descriptive name maybe be: startLocations?
Previously, we also only had Nodes in NodeLayer, but with this change we also have startLocations so NodeLayer name is no longer so descriptive. We could maybe change this to MarkerLayer, or something similar. Open for suggestions.

PR #174, opened 2018-09-10
165 routepath highlights

ahjyrkia:
https://github.com/HSLdevcom/jore-map-ui/issues/165


src/components/map/RouteLayer.tsx
```diff
@@ -34,7 +47,47 @@ export default class RouteLayer extends Component<RouteLayerProps> {
...
         }
+        return false;
+    }
+
+    private mouseHoverIn(internalId: string) {
+        const hoveredPolylines = this.state.hoveredPolylines;
+        hoveredPolylines.push(internalId);
+        this.setState({
+            hoveredPolylines,
+        });
```
kronken: This could be slightly shorter with:

```
private mouseHoverIn(internalId: string) {
    this.setState({
        hoveredPolylines: this.state.hoveredPolylines.concat(internalId),
    });
}
```

src/components/map/RouteLayer.tsx
```diff
@@ -34,7 +47,47 @@ export default class RouteLayer extends Component<RouteLayerProps> {
...
+
+    private isSelected(internalId: string) {
+        if (this.state.selectedPolylines.includes(internalId) ||
+            this.state.hoveredPolylines.includes(internalId)) {
+            return true;
         }
+        return false;
+    }
+
+    private mouseHoverIn(internalId: string) {
```
kronken: Could we rename this function to setLineAsHovered() or similar

src/components/map/RouteLayer.tsx
```diff
@@ -34,7 +47,47 @@ export default class RouteLayer extends Component<RouteLayerProps> {
...
+        const hoveredPolylines = this.state.hoveredPolylines;
+        hoveredPolylines.push(internalId);
+        this.setState({
+            hoveredPolylines,
+        });
+    }
+
+    private mouseHoverOut = () => {
```
kronken: Similarly, a more descriptive name would maybe be: clearHoveredLines()

src/components/map/RouteLayer.tsx
kronken: But this is just a tip, the current code works equally well, and might well be more clearer. So, do as you want.

PR #195, opened 2018-09-17
186 submenu for toolbar

### LINKED ISSUE #186: Submenu for toolbar
Enable submenu in toolbar, lets say we click the "lisää solmu" button. Submenu could then popup and let the user chose which kind of solmu to add

kronken:
Closes #186


kronken:
@Divinare, your suggestion makes sense, and I think it might be a good improvement. However, I don't think it relates to this task. This task was only done to use the underlying logic to create the different subviews that we will need. I created a follow up task in our backlog, which could implement these suggestions and use a so called toolbar button factory. https://github.com/HSLdevcom/jore-map-ui/issues/203

src/components/map/toolbar/toolbarSubmenus.tsx
```diff
@@ -0,0 +1,23 @@
...
+
+const submenus = {
+    edit: (
+        <div>
+            Voit nyt raahata solmut ja pysäkit kartalla.
+        </div>
+    ),
+};
+
+const getSubmenu = (tool: ToolbarTools): JSX.Element | null => {
```
Divinare: I would make this as toolHelp.tsx react component with render function returning:
```
render() {
return (
<div>
       Voit siirtää pysäkkejä tai risteyksiä raahaamalla niitä kartalla.
</div>
)}
```

Divinare: I would make this as toolHelp.tsx react component with render function returning:
```
render() {
if(this.props.tool != ToolbarTool.Edit) return null;

return (
<div>
       Voit siirtää pysäkkejä tai risteyksiä raahaamalla niitä kartalla.
</div>
)}
```

Later if we decide to have some other things in this menu, we can make a wrapper and put things inside, for example:
```
<wrapper>
    <ToolHelp />
   <SnapMenu />
</wrapper>
```

src/components/map/toolbar/Toolbar.tsx
```diff
@@ -1,16 +1,43 @@
...
+                            <ToolbarNetworkButtons />
+                        }
+                    </div>
+                    <div className={s.toolbar}>
+                        <ToolbarCommonButtons />
+                    </div>
+                </div>
+                { submenu !== null &&
+                    <div className={s.toolbarSubMenu}>
+                        {
```
Divinare: How about:
```
{ toolbarStore.activeTool &&
    <ToolHelp
         activeTool=toolbarStore.activeTool 
    />
}
```

Return null inside ToolHelp if activeTool != ToolbarTools.Edit:

kronken:
Review changes done, and all enums set as singular

PR #197, opened 2018-09-17
194 prevent crashing on network failure

### LINKED ISSUE #194: Add error on failed initial request on fetched lines. Currently, the whole site breaks if the backend is down.

kronken:
Closes #194

### DIFF
--- a/src/components/sidebar/SearchResults.tsx
+++ b/src/components/sidebar/SearchResults.tsx
@@ -60,10 +60,9 @@ class SearchResults extends React.Component<ISearchResultsProps, ISearchResultsS
 
     async queryAllLines() {
         this.setState({ isLoading: true });
-        try {
-            this.props.lineStore!.setAllLines(await LineService.getAllLines());
-        } catch (err) {
-            // TODO: show error on screen that the query failed
+        const lines = await LineService.getAllLines();
+        if (lines !== null) {
+            this.props.lineStore!.setAllLines(lines);
         }
         this.setState({ isLoading: false });
     }
--- a/src/services/lineService.ts
+++ b/src/services/lineService.ts
@@ -3,9 +3,10 @@ import apolloClient from '../util/ApolloClient';
 import LineFactory from '../factories/lineFactory';
 import NotificationType from '../enums/notificationType';
 import NotificationStore from '../stores/notificationStore';
+import { ILine } from '../models';
 
 export default class LineService {
-    public static async getAllLines() {
+    public static async getAllLines(): Promise<ILine[] | null> {
         try {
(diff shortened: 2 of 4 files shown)


src/components/sidebar/SearchResults.tsx
```diff
@@ -60,10 +60,9 @@ class SearchResults extends React.Component<ISearchResultsProps, ISearchResultsS
 
     async queryAllLines() {
         this.setState({ isLoading: true });
-        try {
-            this.props.lineStore!.setAllLines(await LineService.getAllLines());
-        } catch (err) {
-            // TODO: show error on screen that the query failed
+        const lines = await LineService.getAllLines();
+        if (lines !== null) {
+            this.props.lineStore!.setAllLines(lines);
```
kronken: Error handling is already handled at the service layer, so we can basically just silently break if there is no result

src/services/lineService.ts
```diff
@@ -3,21 +3,22 @@ import apolloClient from '../util/ApolloClient';
...
+    public static async getAllLines(): Promise<ILine[] | null> {
         try {
             const { data }:any = await apolloClient.query({ query: getLinjas });
             return LineFactory.linjasToILines(data.allLinjas.nodes);
         } catch (err) {
             NotificationStore.addNotification(
                 { message: 'Linjan haku ei onnistunut.', type: NotificationType.ERROR },
             );
-            return err;
+            return null;
```
kronken: My opinion is that is it better to return null if request failed, and we already handled the error. Otherwise the caller don't know what to fail on.

src/services/routeService.ts
```diff
@@ -14,42 +12,37 @@ export interface IMultipleRoutesQueryResult {
...
-        } catch (err) {
-            // tslint:disable-next-line
-            console.error(err);
-            NotificationStore.addNotification({
-                message: 'Reitin haku ei onnistunut.',
-                type: NotificationType.ERROR,
-            });
-            return {
-                nodes: [],
-            };
```
kronken: Again, we don't need to handle the error, since the error is already handled on the service layer.

src/services/routeService.ts
```diff
@@ -14,42 +12,37 @@ export interface IMultipleRoutesQueryResult {
...
+        return null;
     }
 
     public static async fetchRoute(routeId: string): Promise<IRoute | undefined> {
-        return (await RouteService.runFetchRouteQuery(routeId)).route;
+        const routeResult = await RouteService.runFetchRouteQuery(routeId);
+        if (routeResult !== null) {
+            return routeResult.route;
+        }
+        return undefined;
```
kronken: In case of error, the service will return null. The service has already handled the error, so we can silently ignore.

PR #199, opened 2018-09-17
194 prevent crashing on network failure

### LINKED ISSUE #194: Add error on failed initial request on fetched lines. Currently, the whole site breaks if the backend is down.

kronken:
Closes #194

### DIFF
--- a/src/services/lineService.ts
+++ b/src/services/lineService.ts
@@ -27,7 +27,7 @@ export default class LineService {
             NotificationStore.addNotification(
                 { message: 'Linjan haku ei onnistunut.', type: NotificationType.ERROR },
             );
-            return err;
+            return null;
         }
     }
 }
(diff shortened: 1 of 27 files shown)


src/services/lineService.ts
```diff
@@ -3,21 +3,22 @@ import apolloClient from '../util/ApolloClient';
...
+    public static async getAllLines(): Promise<ILine[] | null> {
         try {
             const { data }:any = await apolloClient.query({ query: getLinjas });
             return LineFactory.linjasToILines(data.allLinjas.nodes);
         } catch (err) {
             NotificationStore.addNotification(
                 { message: 'Linjan haku ei onnistunut.', type: NotificationType.ERROR },
             );
-            return err;
+            return null;
```
kronken: My opinion is that is it better to return null instead of any typed error, since we already have handled the error and showed a notification. Otherwise, the caller don't know what to expect.

PR #224, opened 2018-09-24
203 refactoring map buttons

### LINKED ISSUE #203: Refactor mapToolbarButton
Move toolbar/ToolbarButton -> components/controls/mapControlButton.tsx

Rename:
components/controls/toggleButton.tsx -> routePathToggleButton.tsx

neodyymi:
resolves #203


src/components/map/mapControls/MapControlButton.tsx
```diff
@@ -1,6 +1,6 @@
 import React, { Component } from 'react';
 import classnames from 'classnames';
-import * as s from './toolbarButton.scss';
+import * as s from './mapControlButton.scss';
 
 interface ToolbarButtonProps {
```
Divinare: MapControlButtonButtonProps

src/components/shared/styles/common.scss
```diff
@@ -41,3 +41,14 @@ $leafletControlBackgroundHover: rgba(255,255,255,0.5);
 
 $sidebarNarrow: 400px;
 $sidebarWide: 500px;
+
+%soloControl {
```
Divinare: I wouldn't put this in common.css because it is too specific.

If "copy-pasted" css is a problem, make a "mapControl.scss" file for those mapControl buttons. You could have both fullscreenControl.scss and mapControl.scss or just mapControl.scss. Imo here won't be that many styles so only a mapControl.scss shouldn't be a problem.

PR #246, opened 2018-10-02
Set better value for max zoom, 23 is way too close

### LINKED ISSUE #245: Maxzoom is not working

kronken:
Closes #245

### DIFF
--- a/src/components/map/Map.tsx
+++ b/src/components/map/Map.tsx
@@ -155,7 +155,7 @@ class LeafletMap extends React.Component<IMapProps, IMapState> {
                             `
                         }
                         baseLayer={true}
-                        maxZoom={23}
+                        maxZoom={19}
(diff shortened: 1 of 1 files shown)


neodyymi:
6 seems to be really far.. Something like 9-10 would feel better imo

src/components/map/Map.tsx
```diff
@@ -154,7 +154,7 @@ class LeafletMap extends React.Component<IMapProps, IMapState> {
                             `
                         }
                         baseLayer={true}
-                        maxZoom={23}
+                        maxZoom={19}
                         minZoom={6}
```
neodyymi: Should we change this to 9 or 10 while we're at it?

kronken: nah

Issue #241, opened 2018-10-02
Fix dropdown


Divinare:
* Fix bug when scrolling down if scrolldown is visible
* Open dropdown when clicking dropdown

* See comments for thoughts

neodyymi:
I think these should be form items like <select> rather than weird and complex implementations they are now

PR #279, opened 2018-10-08
dropdown fix & fixed input and select generic styles

Divinare:
Resolves: #241 

About the PR:
* Dropdown uses now select & option HTML elements
* Added generic input and select styles


src/components/controls/Dropdown.tsx
```diff
@@ -1,9 +1,9 @@
 import * as React from 'react';
 import * as s from './dropdown.scss';
-import DownArrow from '../../icons/downArrow';
 
 interface IDropdownState {
-    isOpen: boolean;
+    selectedName?: string;
+    selectedValue?: string;
```
neodyymi: Why are these separate? It seems like selectedName is never used.

src/components/controls/Dropdown.tsx
```diff
@@ -17,67 +17,42 @@ class Dropdown extends React.Component
     constructor(props: any) {
         super(props);
         this.state = {
-            isOpen: false,
+            selectedName: undefined,
+            selectedValue: undefined,
```
neodyymi: Why are they undefined when selected is a mandatory prop.

src/components/controls/dropdown.scss
```diff
@@ -1,53 +1,24 @@
...
-    }
+.dropdownView {
+    height: 43px; /* TODO: fix this hack to get dropdown height the same as input field height */
+    margin-top: 5px;
+
+    background:
+        linear-gradient(45deg, transparent 50%, $busBlue 50%),
+        linear-gradient(135deg, $busBlue 50%, transparent 50%);
+    background-position:
+        calc(100% - 21px) calc(16px + 2px),
```
neodyymi: 16+2 = 18

src/components/sidebar/linkView/LinkView.tsx
```diff
@@ -121,13 +121,11 @@ class LinkView extends React.Component<ILinkViewProps, ILinkViewState> {
                         />
                     </div>
                     <div className={s.inputContainer}>
```
neodyymi: `<InputContainer>` and className={s.inputContainer} seems confusing..

neodyymi: Could we just have these dropdowns include the inputcontainer part inside them? With an optional label for the cases where we need "KUTSU-/JÄTTÖ-...."

src/components/sidebar/linkView/LinkView.tsx
```diff
@@ -170,37 +166,31 @@ class LinkView extends React.Component<ILinkViewProps, ILinkViewState> {
...
-                        <Dropdown
-                            onChange={this.onChange}
-                            items={['Kyllä']}
-                            selected={'Ei'}
-                        />
-                    </div>
+                    <Dropdown
+                        onChange={this.onChange}
+                        items={['Kyllä']}
+                        selected={'Ei'}
```
neodyymi: How can 'Ei' be selected if items doesn't contain it?

src/components/sidebar/linkView/LinkView.tsx
```diff
@@ -121,13 +121,11 @@ class LinkView extends React.Component<ILinkViewProps, ILinkViewState> {
                         />
```
neodyymi: Why sometimes use `<InputContainer>` and other times `<div>` and `<input>`?

src/components/controls/Dropdown.tsx
Divinare: If it is not undefined, if I set it as '', there is a warning at console:
Warning: Select elements must be either controlled or uncontrolled (specify either the value prop, or the defaultValue prop, but not both). Decide between using a controlled or uncontrolled select element and remove one of these props

src/components/sidebar/linkView/LinkView.tsx
Divinare: We could have label for dropdown container but then we should use labels for dropdowns as well

Divinare: Made a new card to implement this fix: https://github.com/HSLdevcom/jore-map-ui/issues/283

src/components/sidebar/linkView/LinkView.tsx
Divinare: Not part of this PR. This is only for fixing dropdown. See card #283

src/components/sidebar/linkView/LinkView.tsx
Divinare: This isn't the final solution yet, only a static page. And this is how it was before

Divinare:
Review fixes done

src/components/controls/Dropdown.tsx
neodyymi: Could it not be set as `props.selected` ?

src/components/controls/Dropdown.tsx
neodyymi: fixed in commit b1d78c7

src/components/sidebar/linkView/LinkView.tsx
neodyymi: Sounds good

src/components/sidebar/linkView/LinkView.tsx
neodyymi: 👍

PR #286, opened 2018-10-09
235 setup for deployment

### LINKED ISSUE #235: Setup UI and backend for deployment

kronken:
Closes #235 together with backend PR: https://gitlab.hsl.fi/jore/jore-map-backend/merge_requests/9


Dockerfile
```diff
@@ -3,6 +3,12 @@ FROM node:10-alpine
 WORKDIR /build
 COPY . ./
 
+ARG BACKEND_API_URL
+ENV API_URL=${BACKEND_API_URL}
```
Divinare: How about: BACKEND_API_URL=${BACKEND_API_URL} 

same with GEOSERVER_API_URL

README.md
```diff
@@ -27,7 +27,12 @@ Runs at http://localhost:3000/
 ### Build and run in Docker container
 
 ```bash
+$ # Specify location of API:s
+$ docker build --build-arg BACKEND_API_URL=${API_URL} --build-arg BACKEND_GEOSERVER_URL=${GEOSERVER_URL} --tag=$DOCKER_IMAGE .
```
Divinare: Add an example of API_URL and GEOSERVER_URL?

src/components/map/Map.tsx
```diff
@@ -90,6 +97,7 @@ class LeafletMap extends React.Component<IMapProps, IMapState> {
     }
 
     private fitBounds(bounds: L.LatLngBoundsExpression) {
+        this.getMap().invalidateSize();
```
Divinare: Was this needed for something?

src/components/map/layers/NetworkLayers.tsx
```diff
@@ -28,13 +22,15 @@ interface INetworkLayersProps {
     networkStore?: NetworkStore;
 }
 
+function getGeoServerUrl(layerName: string) {
+    const GEOSERVER_URL = process.env.GEOSERVER_URL || 'http://localhost:8080/geoserver';
+    // tslint:disable-next-line:max-line-length
+    return `${GEOSERVER_URL}/gwc/service/tms/1.0.0/joremapui%3A${layerName}@EPSG%3A900913@pbf/{z}/{x}/{y}.pbf`;
```
Divinare: You forgot to include this very important comment :P
// TODO: Don't hardcode	
// Will be fixed in: https://github.com/HSLdevcom/jore-map-ui/issues/235

travis-build.sh
```diff
@@ -5,12 +5,15 @@ ORG=${ORG:-hsldevcom}
 DOCKER_TAG=${TRAVIS_BUILD_NUMBER:-latest}
 DOCKER_IMAGE=$ORG/jore-map-ui:${DOCKER_TAG}
 DOCKER_IMAGE_LATEST=$ORG/jore-map-ui:latest
+DOCKER_IMAGE_DEV=$ORG/jore-map-ui:dev
```
Divinare: What is the difference between :dev and :latest?

typings.d.ts
```diff
@@ -0,0 +1,7 @@
+declare module '*.scss' {
+  interface IClassNames {
+    [className: string]: string
+  }
+  const classNames: IClassNames;
+  export = classNames;
+}
```
Divinare: Does this have to be in project root?

Dockerfile
kronken: I tried this, it doesn't work. I wanted it to work, It doesn't. I don't know why.

README.md
kronken: I could add the dev servers url, true.

src/components/map/Map.tsx
kronken: Yes, I can add comment

src/components/map/layers/NetworkLayers.tsx
kronken: Well, this is #235, I didn't really think it was hardcoded anymore. I guess it is somewhat hardcoded, but '/gwc/service/tms/1.0.0/joremapui%3A${layerName}@EPSG%3A900913@pbf' is just declaring the type of tiles we want. This functionality is connected to other things in this file, for example that we specify we read the tiles in tms format

travis-build.sh
kronken: We usually have the practise of tagging images based on where they are used. :Latest is just a tag for the latest image.

typings.d.ts
kronken: Maybe not, I don't know actually. The example in the documentation had it in the root, but we could try to move it if you have a better location for it

Dockerfile
Divinare: How about:
ENV BACKEND_API_URL=${_BACKEND_API_URL}?

README.md
kronken: Fixed

src/components/map/Map.tsx
kronken: fixed

Dockerfile
kronken: These are the same variables that are used when building the docker image.
I don't think that the command: docker build --build-arg _BACKEND_API_URL="asd" .
complies to normal env variable standards

src/components/map/layers/NetworkLayers.tsx
Divinare: Ah yes right, I didnt notice this was #235 :P I agree it is less hard coded now. Part of it still is, but maybe it is not worth fixing to a separate file

typings.d.ts
Divinare: The location depends on which file uses this? There is also images.d.ts, it should be moved too.

Dockerfile
Divinare: Didn't quite understand why API_URL=${BACKEND_API_URL} works but ENV BACKEND_API_URL=${_BACKEND_API_URL} doesn't (added an extra _ to make those variables different from each other).

typings.d.ts
kronken: I can try to move them to a common place such as /typings/ or similar

Divinare: Okay, do that and my review is ok!

PR #291, opened 2018-10-09
202 route path info in list

### LINKED ISSUE #202: Edit how each routePath is shown at /routes/
* Show name
    * Tikkurila-Lentoasema
    * Lentoasema-Tikkurila
* Show both timestamps "beginning - end"
* Cut hours off (show only day)

https://github.com/HSLdevcom/jore-map-ui/pull/291

kronken:
Closes #202 

screen shot 2018-10-09 at 16 03 07


src/components/sidebar/routesView/RouteShow.tsx
```diff
@@ -82,7 +87,23 @@ class RouteShow extends React.Component<IRouteShowProps> {
                     key={routePath.internalId}
                 >
                     <div className={s.toggleTitle}>
-                        Suunta {routePath.direction}
+                        <div className={s.toggleTitlePrimary}>
```
Divinare: Nit: I would rename these as
toggle -> routePathContainer
toggleTitle -> routePathInfo
toggleTitlePrimary -> routePathTitle

src/components/sidebar/routesView/RouteShow.tsx
```diff
@@ -82,7 +87,23 @@ class RouteShow extends React.Component<IRouteShowProps> {
...
+                            {`${routePath.originFi}-${routePath.destinationFi}`}
+                        </div>
+                        <div>
```
Divinare: How about using:
s.routePathDate

routePathDate {
    font-size: $smallFontSize
    padding-left: 10px;
}

src/components/sidebar/routesView/RouteShow.tsx
```diff
@@ -25,8 +26,13 @@ interface IRouteShowProps {
...
 class RouteShow extends React.Component<IRouteShowProps> {
-    constructor(props: IRouteShowProps) {
-        super(props);
+    async componentDidMount() {
+        this.props.route.routePaths.forEach((routePath, index) => {
+            // Make two first route paths visible by default
+            if (index < 2) {
+                this.props.routeStore!.toggleRoutePathVisibility(routePath.internalId);
+            }
+        });
```
Divinare: This is a hack imo. I think its not the component's responsibility to set first two visible. I would set the property visible for the first two routePaths when they are put to the store for the first time if < 2 routePaths are visible.

Nit: also this isn't in the scope of this task.

Edit: hmm, it was factory's job before... but maybe thats not the correct place either...

src/components/sidebar/routesView/routeShow.scss
```diff
@@ -45,7 +45,11 @@
 
         .toggleTitle {
             display: inline-block;
-            width: 110px;
+            width: 270px;
```
Divinare: Static width is not a good solution (I know it was before a static width as well). But here is the point: if it wasn't a static width, you wouldn't have had to change it here again. So avoid using static widths :P (of course there are exceptions...)

src/models/IRoutePath.ts
```diff
@@ -12,4 +12,8 @@ export default interface IRoutePath {
     endTime: Date;
     lastModified: Date;
     routePathLinks: IRoutePathLink[];
+    originFi: string;
+    originSe: string;
+    destinationFi: string;
+    destinationSe: string;
```
Divinare: destinationSe and originSe not being used. Why do we query them in if we don't use them?

src/components/sidebar/routesView/RouteShow.tsx
kronken: In my opinion this belongs to the component. Before it was, as you noticed in the factory, but the factory should just create the object for us. In my opinion the factory should not need to know even what the visible thing means, it should just deliver a object from Jore.
I think it is the components responsibility to set this variable, in my opinion the component is the only module that actually needs to know what the variable even does. This is because the variable is only used in this specific component, and the visible property is then reflected forward on the map.

src/components/sidebar/routesView/RouteShow.tsx
kronken: true, fixing

src/components/sidebar/routesView/RouteShow.tsx
kronken: nice, fixing

src/components/sidebar/routesView/routeShow.scss
kronken: I guess why this previously was static is that the toggle button wanted to be floating at a specific point. This change sets the toggle to be in the right side, so I could indeed see if we could remove this static thing.

src/models/IRoutePath.ts
kronken: I like swedish

kronken: I'm kidding, I'll remove it

src/components/sidebar/routesView/routeShow.scss
kronken: this is fixed

Divinare:
Review ok now

Issue #210, opened 2018-09-18
Route index page


Divinare:
* is there something we can do with this? Otherwise just redirect to /root/

I would say we just redirect the user to index

PR #292, opened 2018-10-09
routes page without query params redirects to root

Divinare:
Resolves: #210


src/components/sidebar/Sidebar.tsx
```diff
@@ -35,6 +36,12 @@ interface ILinelistState {
 @inject('sidebarStore', 'routeStore', 'searchStore')
 @observer
 class Sidebar extends React.Component<ISidebarProps, ILinelistState> {
+
+    private renderRoutesView = () => {
+        const search = this.props.history.location.search;
+        return search === '' ?  <Redirect to='/' /> : <RoutesView />;
+    }
```
kronken: The history things could be skipped if we utilize the navigator for this:
```
private renderRoutesView = () => {
    return !navigator.getQueryParam(QueryParams.routes)
        ? <Redirect to='/' /> : <RoutesView />;
}
```

kronken: I would had put this logic inside the route view, and there just tell the navigator to navigate to '/'.

Divinare:
Added todo to rename enum QueryParam. It could be done here: https://github.com/HSLdevcom/jore-map-ui/issues/137

Review comment fixed.

src/routing/queryParams.ts
```diff
@@ -1,3 +1,4 @@
+// TODO: rename QueryParams https://en.wikipedia.org/wiki/Query_string
```
kronken: Query string is the whole part of the URL string containing Query Parameters. A single label value pair is called a query parameter, so I think the current naming is OK.

PR #293, opened 2018-10-10
Showing warning and painting node red

### LINKED ISSUE #289: Fix: Node type not supported
http://jore-map-dev.hsldev.com/routes/?
routes[0]=0039

This shouldn't happen, console:

react-dom.production.min.js:188 Error: NodeType not supported: not_found!
    at e.getIcon (NodeLayer.tsx:74)
    at NodeLayer.tsx:116
    at Array.map (<anonymous>)
    at e.render (NodeLayer.tsx:115)
    at X (mobx.module.js:618)
    at index.module.js:758
    at dt (mobx.module.js:1089)
    at t.track (mobx.module.js:1608)
    at e.n [as render] (index.module.js:753)
    at Pi (react-dom.production.min.js:167)

kronken:
Closes #289 
screen shot 2018-10-10 at 14 20 37


Divinare:
Our notification store currently filters away error messages that have the same length. Should this be fixed here or create a new card?

I mean this line of code should be fixed:
if (this._notifications.filter(n => n.message === notification.message).length === 0) {

src/components/map/layers/NodeLayer.tsx
```diff
@@ -30,9 +33,13 @@ enum color {
     STOP_BORDER_COLOR_SELECTED = '#007ac9',
     STOP_FILL_COLOR = '#FFF',
     STOP_FILL_COLOR_SELECTED = '#007ac9',
+    UNKNOWN_BORDER_COLOR = '#ff0000',
+    UNKNOWN_BORDER_COLOR_SELECTED = '#ff0000',
+    UNKNOWN_FILL_COLOR = '#000',
+    UNKNOWN_FILL_COLOR_SELECTED = '#000',
```
Divinare: If nodeType is unknown, should it fillcolor be white and border color red so that it would be more visible?

Divinare:
Also I wish that our INotification would have a geometry property so that you could optionally pass the geometry of invalid feature to the notificationStore and user would be able to fit bounds of the map to the invalid feature.

src/components/map/layers/NodeLayer.tsx
```diff
@@ -71,7 +89,11 @@ export default class NodeLayer extends Component<MarkerLayerProps> {
             break;
         }
         default: {
-            throw new Error(`NodeType not supported: ${node.type}!`);
+            this.props.notificationStore!.addNotification({
+                message: `Solmu id: '${node.id}', tyyppi '${node.type}' on virheellinen`,
```
Divinare: How about format "Solmun 123123 (id) tyyppi puuttuu."

And is this really something that should be notified? Ask customer perhaps.

src/components/map/layers/NodeLayer.tsx
```diff
@@ -57,6 +64,17 @@ export default class NodeLayer extends Component<MarkerLayerProps> {
...
     }
 
+    private getUnknownNodeMarkerHtml = (isSelected: boolean) => {
+        const borderColor = isSelected ?
+            color.UNKNOWN_BORDER_COLOR_SELECTED : color.UNKNOWN_BORDER_COLOR;
+        const fillColor = isSelected ?
+            color.UNKNOWN_FILL_COLOR_SELECTED : color.UNKNOWN_FILL_COLOR;
+        return `<div
+            style="border-color: ${borderColor}; background-color: ${fillColor}"
+            class=${s.nodeContent}
```
Divinare: I see this inline html javascript 3 x times ---> make a function to say it only once

src/components/map/layers/NodeLayer.tsx
Divinare: Other solutions: a yellow/red exclamation mark next to that [ i ] button in sidebar (routeShow). Clicking that [ ! ] button you could see a small window of all invalid properties or smth...

src/components/map/layers/NodeLayer.tsx
Divinare: Also hmm, we could just use class to declare all of this just like in crossroadCircle and stopCircle.

`<div class=${s.nodeContent s.unknownNode}></div>`

.unknownNode {
   color: ...
   fillColor: ... or is it just fill: ?
}

So instead of saying:
html = this.getUnknownNodeMarkerHtml(isSelected);

you should say:
html = `<div class=${s.nodeContent s.unknownNode}></div>`

src/components/map/layers/NodeLayer.tsx
kronken: We can't really say that it is missing, since we don't have that information. The only thing we can say that it is invalid, since it cannot be mapped to the mappings we have in the UI.
And I think it is pretty clear that this should be visible to the users, finding errors such as these is one important improvement over the old program.

src/components/map/layers/NodeLayer.tsx
```diff
@@ -63,15 +40,19 @@ export default class NodeLayer extends Component<MarkerLayerProps> {
...
-            html = this.getNodeCrossroadMarkerHtml(isSelected);
+            html = this.getMarkerHtml(isSelected ? s.crossroadMarkerHighlight : s.crossroadMarker);
             break;
         }
         default: {
-            throw new Error(`NodeType not supported: ${node.type}!`);
+            this.props.notificationStore!.addNotification({
+                message: `Solmun (id: '${node.id}') tyyppi on virheellinen`,
+                type: notificationType.WARNING,
+            });
```
Divinare: This addNotification brings an error msg to the console:

index.js:2178 Warning: Cannot update during an existing state transition (such as within `render` or another component's constructor). Render methods should be a pure function of props and state; constructor side-effects are an anti-pattern, but can be moved to `componentWillMount`.

Issue #322, opened 2018-10-16
Differentiate stops that are not in use on routepath


neodyymi:
Some stops on a routepath are not in use. Visualize this on map.

PR #332, opened 2018-10-22
322 differentiate disabled stops

ahjyrkia:
closes https://github.com/HSLdevcom/jore-map-ui/issues/322


src/factories/routePathLinkFactory.ts
```diff
@@ -10,7 +10,11 @@ class RoutePathLinkFactory {
     public static createRoutePathLink = (routePathLinkNode: any): IRoutePathLinkResult => {
         const nodes = [];
         if (routePathLinkNode.solmuByLnkalkusolmu) {
-            nodes.push(NodeFactory.createNode(routePathLinkNode.solmuByLnkalkusolmu));
+            const node = NodeFactory.createNode(routePathLinkNode.solmuByLnkalkusolmu);
+            if (routePathLinkNode.relpysakki === 'E') {
+                node.disabled = true;
+            }
+            nodes.push(node);
```
kronken: Correct me if I'm wrong, but won't this cause problems when two routes use the same stop? What if the stop is disabled on one of the routes, but not the other. This code may mark one of the stops as disabled, but leave the other stop from the other routes active. If looking at function that calls this code, we can see that it collect all nodes, and remove duplicates by looking at ID. The duplicate function wont recognise difference in activity. This leads to a problem that where we have a race condition that sets the outcome. If the active stop is before the disabled stop in the list, it will lead the merged stop to be active. However, if the disabled stop is before, it will lead the merged stop to be disabled.

src/components/map/Map.tsx
```diff
@@ -122,14 +122,24 @@ class LeafletMap extends React.Component<IMapProps, IMapState> {
         this.getMap().setView(latLng, 17);
     }
 
+    private setDisabledNodeIds = (links: IRoutePathLink[]) => {
```
Divinare: Why have setDisabledNodeIds at Map.tsx? 

What about having a method at RouteLayer

setRouteOnTop() {
    setHoveredPolylines();
    setDisabledNodeIds();
}

Divinare: Also is there a way to go without forceUpdate()?

src/factories/routePathFactory.ts
```diff
@@ -50,6 +50,7 @@ class RoutePathFactory {
 
         return {
             routePath,
+            // TODO Determine duplicates by coordinates.
```
Divinare: Why this TODO comment? Can node's have the same ID?

I think there can be a case where 2 nodes could be on top of each other so determing by coordinates might not work either

src/models/IRoutePathLink.ts
```diff
@@ -4,4 +4,5 @@ export default interface IRoutePathLink {
     startNodeId: string;
     endNodeId: string;
     orderNumber: number;
+    relPysakki: string;
```
Divinare: relPysakki? Why?

src/components/map/Map.tsx
ahjyrkia: Yes. It is no longer needed after moving setDisabledNodeIds to RouteLayer.

PR #334, opened 2018-10-23
333 Subtask to Saving forms

### LINKED ISSUE #333: Subtask to #299, Setting up RoutePathView to allow saving functions
https://github.com/HSLdevcom/jore-map-ui/pull/334

kronken:
Closes #333 

Had to do some setting up of RoutePath in order to be able to continue with saving functions.
Changes include:
- Pass ID to view when opening RoutePathView
- Based on ID, fetch RoutePath
- Take fetched values and insert them in RoutePathView
- Add initial 'save' button
- Add 'isDirty' state to track if change is made


src/routing/queryParams.ts
```diff
@@ -1,6 +1,9 @@
 // TODO: rename QueryParams https://en.wikipedia.org/wiki/Query_string
 enum QueryParams {
     routes= 'routes',
+    routeId= 'routeid',
+    startTime= 'starttime',
+    direction= 'direction',
```
Divinare: Why are startTime and direction params added instead of routePathId?

I know that routePathId is generated at our frontend.

I see 2 ways of fixing this:

Option 1: Make routePathId format so that our frontend can decode it. So don't use hash when gererating it.

Option 2: What if our backend generated routePathId and we could query a specific routePath from our backend with that id?

src/components/sidebar/routePathView/RoutePathView.tsx
```diff
@@ -16,55 +25,101 @@ class RoutePathView extends React.Component<IRoutePathViewProps, IRoutePathViewS
     constructor(props: any) {
         super(props);
         this.state = {
-            isEditingDisabled: false,
+            isEditingDisabled: true,
+            isDirty: false,
```
Divinare: This doens't say to me what this variable means.

isSaveDisabled?

src/components/sidebar/routePathView/NewRoutePathView.tsx
```diff
@@ -1,33 +1,71 @@
...
-            <RoutePathViewForm
-                isEditingDisabled={false}
-            />
-            <div>
-                <div className={s.sectionDivider} />
-                <div className={s.padding} />
+            <div className={s.section}>
+                <RoutePathViewForm
+                    isEditingDisabled={false}
+                    onEdit={this.onChange}
```
Divinare: Why not say onChange={this.onChange}?

src/components/sidebar/InputContainer.tsx
```diff
@@ -7,10 +7,18 @@ interface IInputProps {
     placeholder?: string;
     className?: string;
     disabled?: boolean;
+    onChange?: Function;
+    value?: string;
 }
 
 class InputContainer extends React.Component<IInputProps> {
     public render(): any {
+        const onChange = (e: React.FormEvent<HTMLInputElement>) => {
```
Divinare: I'm not sure if this is the correct way of implementing input field binding. Smaller cards would be preferred.

src/routing/queryParams.ts
kronken: I don't see this as a problem. The primary keys of routePaths in Jore are routeId, startTime and direction. When fetching data from the backend, we need to use these primary keys. As you told, we cannot use any hash, but I don't see the need for encoding these primary keys, and later decoding them.
I don't see any problem with setting all these values in the URL, and I think that is a better solution than having some made up ID, that we need to encode/decode.

src/components/sidebar/InputContainer.tsx
kronken: This is how React documentation does it, I'm open to better alternatives.
https://reactjs.org/docs/forms.html

src/components/sidebar/routePathView/RoutePathView.tsx
kronken: In my opinion variable names should not be set to what they affect, the name should be set to what the variable means. This variable tells if the current object opened is dirty. It tells if there have been modifications to the object, and these changes have not been synchronised to the database.

Divinare: Hmm, I see your point why it is "isDirty". Haven't used that term before. How about "hasModifications"?

Divinare: Also generally about the naming: naming with what the variable does/where it affects can also mean at the same time what it means :P

src/routing/queryParams.ts
Divinare: I see. The reason I said about this that now we refer to a routePath with 2 ways, our generated routePathId and routeId+startTime+direction. I wish there was only one way but maybe it would also make our lives more difficult. So maybe i'm fine with this solution too.

src/routing/queryParams.ts
```diff
@@ -1,6 +1,9 @@
 // TODO: rename QueryParams https://en.wikipedia.org/wiki/Query_string
 enum QueryParams {
     routes= 'routes',
+    routeId= 'routeid',
+    startTime= 'starttime',
```
Divinare: nit: startTime

src/components/sidebar/routePathView/RoutePathViewForm.tsx
```diff
@@ -1,216 +1,246 @@
...
+        // TODO
     }
 
-    // TODO
-    public onChange = () => {
+    public onChange = (property: string) => (value: string) => {
+        this.props.onEdit();
+        this.setState({
+            routePath: { ...this.state.routePath, [property]: value },
+        });
```
Divinare: I would call onChange of parent instead.

Then routePath to this form could always be given only through props (no need to have it at state)

Reason: at NewRoutePathView there is save() method that takes the edited routePath data and sends it to the backend.

So rename onEdit() -> onChange()?

src/components/controls/Button.tsx
```diff
@@ -35,16 +36,28 @@ class Button extends React.Component<IButtonProps, {}> {
...
 
     public render(): any {
         return (
             <div
-                className={this.getClassname(this.props.type, this.props.className)}
-                onClick={this.props.onClick}
+                className={
+                    this.getClassname(
+                        this.props.type,
+                        this.props.disabled!,
```
Divinare: Use this.props.disabled at getClassName(), no need to pass here as param

src/components/controls/button.scss
```diff
@@ -60,3 +60,8 @@
 .save:hover {
     background-color: darken($saveButtonGreen, 10);
 }
+
+.disabled {
+    background-color: $disabledGrey !important;
```
Divinare: Don't use !important. Use
```
pointer-events: none;
background-color: $disabledGrey
```
instead

src/components/shared/styles/form.scss
```diff
@@ -4,6 +4,10 @@
     margin-right: 10px;
     padding-top: 10px;
     width:100%;
+
+    .disabled {
+        background-color: $disabledGrey;
```
Divinare: Add pointer-events: none;?

src/components/sidebar/routePathView/NewRoutePathView.tsx
```diff
@@ -17,17 +55,17 @@ class NewRoutePathView extends React.Component{
...
             />
-            <div className={s.padding} />
-            <div className={s.padding} />
-            <RoutePathViewForm
-                isEditingDisabled={false}
-            />
-            <div>
-                <div className={s.sectionDivider} />
-                <div className={s.padding} />
+            <div className={s.section}>
```
Divinare: If you look at the HTML, there is this structure:

```
section
    div
          section
```
which doensn't seem right.

Also that 5 button section doesn't have a section wrapper

src/components/shared/styles/form.scss
```diff
@@ -24,8 +28,12 @@
     font-size: $mediumFontSize;
 }
 
-.sectionDivider {
-    padding-top:20px;
-    border-bottom: 1.5px solid $lightGrey;
-    width: 100%;
-}
+.section {
+    margin: 30px 0;
```
Divinare: We have been using padding instead of margin.

Also what if you removed this whole .section wrapper and added padding-top + padding-bottom to .sectionDivider?

Alternative solution would be to remove .sectionDivider,  add border-bottom, padding-top, padding-bottom to .section and to :last-type-of .section add border-bottom: none

src/components/sidebar/routePathView/NewRoutePathView.tsx
Divinare: Also note: before changing anything see what I proposed at .margin comment. I think there might be no need for this wrapper

src/services/routePathService.ts
```diff
@@ -0,0 +1,58 @@
...
+import moment from 'moment';
+import apolloClient from '~/util/ApolloClient';
+import { IRoutePath } from '~/models';
+import notificationStore from '~/stores/notificationStore';
+import NotificationType from '~/enums/notificationType';
+import RoutePathFactory from '../factories/routePathFactory';
+
+export default class RoutePathService {
+    public static async fetchRoutePath
+        (routeId: string, startDate: moment.Moment, direction: string): Promise<IRoutePath | null> {
```
Divinare: What means startDate: moment.Moment? Shouldn't this be the real startDate of routePath?

src/components/controls/Button.tsx
```diff
@@ -7,13 +7,14 @@ interface IButtonProps {
     type: ButtonType;
     text: string;
     className?: string;
+    disabled?: boolean;
```
neodyymi: does this need to be optional if defaultProps gives it a value?

src/components/shared/styles/form.scss
neodyymi: I sort of like the idea that there's no need for a section divider. Now the sectiondividers are just bolted in and don't really mean anything.. CSS selectors could probably handle this.

src/components/sidebar/routePathView/RoutePathViewForm.tsx
kronken: I noticed this too, when I started to make the continuation task. There I have moved the logic to the parent, I also added a comment there to consider moving the onEdit -> onChange.
https://github.com/HSLdevcom/jore-map-ui/pull/338/files#diff-7303060f0e3fa68e5bcb30f851aaa7d7R79

src/factories/routePathFactory.ts
```diff
@@ -17,43 +17,50 @@ class RoutePathFactory {
...
-                suunta.suuvoimviimpvm,
                 suunta.suusuunta,
             ].join('-'),
         ).toString();
 
-        const routePathLinkResult:IRoutePathLinkResult[]
-        = suunta.reitinlinkkisByReitunnusAndSuuvoimastAndSuusuunta.edges
+        const routePathLinkResult:IRoutePathLinkResult[] | null
+        = suunta.reitinlinkkisByReitunnusAndSuuvoimastAndSuusuunta ?
+            suunta.reitinlinkkisByReitunnusAndSuuvoimastAndSuusuunta.edges
```
neodyymi: Making a little tweak to the query would allow removal of .edges and .node. Not sure which way we want to do it though. Personally I find this edges-node less intuitive than using .nodes

src/services/routePathService.ts
neodyymi: moment.Moment is the type for a timestamp when using moment.js

src/util/nodeHelper.ts
```diff
@@ -47,7 +47,7 @@ export default class NodeHelper {
 
     public static routePathHasStop(routePaths: IRoutePath[], nodeId: string) {
         return routePaths.some(routePath =>
-            routePath.routePathLinks.some(routePathLink =>
+            routePath.routePathLinks!.some(routePathLink =>
```
neodyymi: Does this need to account for a stop not being in use on the routepathlink?

src/components/sidebar/routePathView/RoutePathView.tsx
```diff
@@ -1,12 +1,21 @@
...
+import RoutePathService from '~/services/routePathService';
+import Loader from '~/components/shared/loader/Loader';
 import ViewHeader from '../ViewHeader';
 import RoutePathViewForm from './RoutePathViewForm';
 import * as s from './routePathView.scss';
 
 interface IRoutePathViewState {
     isEditingDisabled: boolean;
+    isDirty: boolean;
+    routePath: IRoutePath | null;
```
neodyymi: Why `| null` instead of `?` ?

src/components/controls/Button.tsx
kronken: Personally, I think the scope of functions is better visualized if we try to pass used as variables as parameters, instead of use them in the function directly from props.

from the 'this.getClassname(a,b,c)' I directly see on what basis the function makes its decision. 
If we would have 'this.getClassname()', we know it return something, but we don't know on what basis.

Note, I don't think any answer to this is wrong or correct, this is just my personal preference.

src/factories/routePathFactory.ts
kronken: Yes, we probably use this at many places, we could refactor them to using simpler form, but maybe not in this task

src/util/nodeHelper.ts
kronken: Yes, If I understand you correctly. The routePathLink data is not needed in routePath view, so I don't fetch it. Which means I need to allow it to be empty

src/components/controls/Button.tsx
kronken: yes

Divinare:
Conflicts should be fixed

Divinare:
Our "reitinSuunta" view's topic has a border bottom, linkView's text area has margin-top.

Also fix conflicts

src/components/controls/Button.tsx
Divinare: How about using this code:
```
                className={
                    classnames(
                        s.button,
                        this.props.className,
                        this.getTypeClass(this.props.type),
                        this.props.disabled ? s.disabled : null,
                    )
                }
```

and remove getClassname method.

Divinare: Fixed already.

Issue #343, opened 2018-10-29
Static colors part 2 - "good" range of colors


Divinare:
Use a static range of colors picked from known good assortments

PR #344, opened 2018-10-29
323 change routes color range to static

### LINKED ISSUE #323: Change color range to static
* Change route colors to no change when toggling them on and off.
* Use a static range of colors picked from known good assortments

Divinare:
Closes #323 

Using a library of static range colors was in the original scope of this task. However there ended up being so much changes so I split this to a new task: #343

About the solution:
* Before we had RouteLayer & RoutePathLayer, now we have RoutePathLinkLayer too
* ColorScales getColorMap() method ensures consistent way of getting the color of a certain routePath


src/components/map/layers/RoutePathLayer.tsx
```diff
@@ -1,47 +1,58 @@
...
+    colors: string[];
+    routePaths: IRoutePath[];
+    fitBounds: (bounds: L.LatLngBoundsExpression) => void;
 }
 
-export default class RoutePathLayer extends Component<RoutePathLayerProps> {
+@inject('sidebarStore')
+@observer
+export default class RouteLayer extends Component<RoutePathLayerProps> {
+    constructor(props: RoutePathLayerProps) {
```
kronken: unnecessary constructor

src/util/colorScale.ts
```diff
@@ -1,7 +1,25 @@
...
 
 class ColorScale {
-    public static getColors(colorCount: number) {
+    public static getColorMap(routes: IRoute[]) {
+        const colorMap = new Map();
+
+        let totalColorCount = 0;
+        routes.forEach((route) => {
+            totalColorCount += route.routePaths.length;
+        });
```
kronken: const totalColorCount = route.reduce((acc,curr) => curr.routePaths.length + acc, 0);

src/util/colorScale.ts
```diff
@@ -1,7 +1,25 @@
...
+        routes.forEach((route) => {
+            const key = route.routeId;
+            const value = colors.splice(0, route.routePaths.length);
+            colorMap.set(key, value);
+        });
+
+        return colorMap;
+    }
+
+    private static getColors(colorCount: number) {
```
neodyymi: I really think we would get better results from a proper range of distinct colours. They're designed from a ux point of view and from the last sprint demo/planning this is what I understood the users wanted - a limited number of well chosen colours.

Divinare: Will be implemented here: https://github.com/HSLdevcom/jore-map-ui/issues/343

Let's discuss/plan this at slack

PR #347, opened 2018-10-30
Store import convention added

Divinare:


Contributions.md
```diff
@@ -31,6 +31,8 @@
 
 ### MobX
 
+* Import store through props always if possible. Use import without props only if needed.
```
neodyymi: `@inject` stores when possible...

PR #348, opened 2018-10-30
highlighting neighbor nodes & handling click of neighbor nodes

### LINKED ISSUE #296: Part 2. Creating new routePath geometry - highlight neighbouring nodes & handling click of neighbor nodes
Blocked by #295 

* Highlight neighbouring nodes to visualize possible directions.

* prevent clicking a crossroad. Also don't highlight crossroads

* clicking neighbor node adds it to the newRoutePathStore

Divinare:
closes #296 

* Added click handling of neighbor nodes to the scope as well.

* Notice that some of the nodes are on top of each other. Don't worry about that in this PR

I edited part 3 card description. Will add routePathLinks & routePathLinkHighlights there.


src/components/map/layers/new/NewRoutePathLayer.tsx
```diff
@@ -14,37 +14,58 @@ interface IRoutePathLayerProps {
...
-
     }
 
     private renderNodes() {
         return this.props.newRoutePathStore!.nodes.map((node, index) => {
             const latLng = L.latLng(node.coordinates.lat, node.coordinates.lon);
             return (
                 <Marker
-                    icon={this.getIcon()}
+                    icon={this.getIcon(false)}
```
kronken: Could try to pass object instead.

this.getIcon({isNeightbor: false})
public getIcon({isNeighbor: boolean})
public getIcon(options)
const { isNeighbor } = options

src/components/map/layers/NetworkLayers.tsx
```diff
@@ -125,23 +119,28 @@ export default class NetworkLayers extends Component<INetworkLayersProps> {
...
         const properties =  clickEvent.sourceTarget.properties;
+        if (properties.soltyyppi !== NodeType.STOP) return;
+
         // TODO: Use factory call instead of service call because
         // geojson / geojsonManual are not found from properties
         NodeService.fetchNode(properties.soltunnus).then((node) => {
             if (node) {
                 this.props.newRoutePathStore!.addNode({
-                    id: node.id,
+                    nodeId: node.id,
```
kronken: It seems we have a bit different coding styles for different objects. My preference is to always set the primary key (id) property called 'id'. So, Node.id, Stop.id, Line.id, Route.id. And it seems this is the way that it is currently done for 4 of our models. 2 of our models use the style that you now use here, for example: (route.routeId, line.lineId). We should use the same style for all models, so, I don't see the reason for changing this to 'nodeId'

src/services/lineService.ts
```diff
@@ -22,6 +23,7 @@ export default class LineService {
         }
     }
 
+    // TODO: rename as fetchLine
     public static async getLine(lintunnus: string): Promise<ILine | null> {
```
kronken: Why TODO for this? Why not just change this already now, getLine and getAllLines are just called at one place respectively. So, there would not be too much work to just replace them.

src/components/map/layers/NetworkLayers.tsx
Divinare: Yes we should use either id or nodeId

PR #362, opened 2018-11-06
304 change log

ahjyrkia:
closes https://github.com/HSLdevcom/jore-map-ui/issues/304


ahjyrkia:
New button under the toolbar

PR #368, opened 2018-11-12
NewRoutePath: showing links & nodes + neighbor nodes & links

### LINKED ISSUE #297: Part 3. Creating new routePath geometry, - routePathLinks & neighbor routePathLink highlighting
Blocked by #361 

Initial work started at branch: 297-createNewRoutePath-highlight-routePathLinks

* Add routePathLinks between added nodes to the store
* Highlight neighbor routePathLinks

Divinare:
closes #297

About the implementation
* Added stops are rendered as green border, crossroads as grey border
* Neighbor nodes / routePathLinks are purple
    * You can only click neighborNodes
* Nodes aren't rendered as disabled / timeAlignmentStop because we don't have this info yet
* Sometimes "routePathLink" is called as link - we know it is routePathLink due to the context "NewRoutePathLinkLayer"

* Things like draw order or hover highlight of neighbor links should be done in other cards. There is also some "double click bug" that happens when u click networkLayer -> should be done in other card


src/components/map/layers/new/NewRoutePathLayer.tsx
```diff
@@ -1,73 +1,126 @@
...
+                color={color}
+                weight={5}
+                opacity={0.8}
+                onClick={isNeighbor ? this.addLinkToRoutePath(link) : null}
+            />
+        );
+    }
+
+    private addLinkToRoutePath = (link: IRoutePathLink) => () => {
+        RoutePathLinkService.fetchLinksWithLinkStartNodeId(link.endNode.id).then((links) => {
```
kronken: async await

Divinare:
Review fixes done

Issue #388, opened 2018-11-19
Editing routePath form is not working at routePath/new


Divinare:
blocked by #298 

* Also consider taking routePath out from NewRoutePathView's state.

Issue #382, opened 2018-11-13
Find advantages of having graphql.ts file that exports reusable graphql query functions


Divinare:
Graphql.ts

```
fetchRoute() {
   const routeQuery = getRouteQuery();
   const queryResult = apollo.run(routeQuery)
   return getExternalRoute(queryResult);
}

getRouteQuery {
   const routePathQuery = getRoutePathQuery()
    return {
        reitunnus
        reinimi
        reinimilyh
        ...
        ${routePathQuery}
    }
}

getRoutePathQuery() { ... }

fetchRoutePath() {
   const routePathQuery = getRoutePathQuery();
   const queryResult = apollo.run(routePathQuery)
   return getExternalRoutePath(queryResult);
}
```

@neodyymi : We should look into this https://www.apollographql.com/docs/react/recipes/static-typing.html

Joe: typing like that could be done but still we need reusable graphql queries

PR #390, opened 2018-11-19
298 new route path linking

### LINKED ISSUE #298: Part 4. Creating new routePath geometry - linking to an existing line and route
Started work at branch: 298-newRoutePath-linking

* at routeList, user can click "luo uusi reitin suunta"
* at routePathView, user can click "luo uusi reitin suunta käyttäen tätä pohjana"

* routeId and linkId are automatically linked to new routePath

* URL is in format: route/:routeId/newRoutePath

Divinare:
closes #298 

About the implementation:
* created two bug tasks #388 #389
* routePathLinks had to be sorted by orderNumber (they were in random order). I also sorted routePaths in routeService
* new routePath is now at routePathStore and also currentRoutePath is at NewRoutePathView's state (maybe that routePath should be taken out of the state, TODO in here -> #388?
* I also edited the information of the current routeId / lineId (as you can see, for example "LINJA: 0033
REITTI: 0033"
* Those graphQL fetched should be implemented in one common graphql.ts file, done in here: #382


src/components/map/layers/new/NewRoutePathLayer.tsx
```diff
@@ -112,6 +112,23 @@ export default class RoutePathLayer extends Component<IRoutePathLayerProps> {
         return this.renderNode(firstNode, link, { isNeighbor: false }, 0);
     }
 
+    async componentDidUpdate() {
```
kronken: Why do we have this in componentDidUpdate, instead of just calling it as a function where it is needed? From my limited understanding of this chain of events, would it be possible to call it after addLinkToRoutePath?

src/components/sidebar/routePathView/NewRoutePathView.tsx
```diff
@@ -4,68 +4,60 @@ import classnames from 'classnames';
...
+interface INewRoutePathViewState {
+    currentRoutePath?: IRoutePath;
+}
+
+@inject('mapStore', 'routeStore', 'routePathStore', 'networkStore')
 @observer
-class NewRoutePathView extends React.Component<INewRoutePathViewProps>{
-    initialRoutePath: IRoutePath;
+class NewRoutePathView extends React.Component<INewRoutePathViewProps, INewRoutePathViewState>{
+    currentRoutePath: IRoutePath;
```
kronken: What is this variable: currentRoutePath? Is this necessary? do we use it?

src/components/sidebar/routePathView/NewRoutePathView.tsx
```diff
@@ -4,68 +4,60 @@ import classnames from 'classnames';
...
-            routePathLinks: [],
-            originFi: '',
-            originSw: '',
-            destinationFi: '',
-            destinationSw: '',
-            routePathShortName: '',
-            routePathShortNameSw: '',
-            modifiedBy: '',
+        this.state = {
+            currentRoutePath: undefined,
```
kronken: I think this could be written shorter with:

this.state = {};

kronken: Since, both ways will make this.state.currentRoutePath to be undefined

src/components/sidebar/routePathView/NewRoutePathView.tsx
```diff
@@ -4,68 +4,60 @@ import classnames from 'classnames';
...
+        });
     }
 
-    // TODO
     public onSave = () => {
+        // TODO
+    }
+
+    componentWillMount() {
+        if (!this.props.routePathStore!.routePath) {
```
kronken: Usually we use function names to describe what a section does. This becomes a problem when we use lifecycle methods, since they don't describe what the functionality does. I would suggest extracting the code in the lifecycle method in to something like initEmptyRoutePath() or similar

src/components/sidebar/routePathView/NewRoutePathView.tsx
```diff
@@ -75,27 +67,46 @@ class NewRoutePathView extends React.Component<INewRoutePathViewProps>{
         this.props.networkStore!.setIsLinksVisible(true);
 
         this.props.routePathStore!.setIsCreating(true);
+
+        // TODO: this should be action call
```
kronken: I also find this lifecycle a bit difficult to understand, would it be possible to move some logic in this to other functions as well?

src/services/lineService.ts
```diff
@@ -17,7 +17,8 @@ export default class LineService {
                 const externalLine = this.getExternalLine(linja);
                 return LineFactory.createLine(externalLine);
             }));
-        } catch (err) {
+        } catch (error) {
+            console.log(error); // tslint:disable-line
```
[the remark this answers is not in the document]
Divinare: Fixed

Divinare:
Joakim's review fixes done ^

src/components/sidebar/routePathView/NewRoutePathView.tsx
neodyymi: Or just adding a comment here could do the trick

src/components/sidebar/routePathView/NewRoutePathView.tsx
```diff
@@ -4,98 +4,111 @@ import classnames from 'classnames';
...
+        const newRoutePath =
+            RoutePathFactory.createNewRoutePath(queryParams.lineId, queryParams.routeId);
 
+        this.props.routePathStore!.setRoutePath(newRoutePath);
     }
 
     componentDidMount() {
-        this.props.mapStore!.setNodeSize(NodeSize.large);
+        this.initStores();
+        this.initCurrentRoutePath();
```
neodyymi: Do this already on constructor as discussed

src/components/sidebar/routePathView/NewRoutePathView.tsx
```diff
@@ -4,98 +4,111 @@ import classnames from 'classnames';
...
     componentWillUnmount() {
         this.props.mapStore!.setNodeSize(NodeSize.normal);
         this.props.routePathStore!.setIsCreating(false);
-
+        this.props.routePathStore!.setRoutePath(null);
     }
 
     public render(): any {
+        const currentRoutePath = this.state.currentRoutePath;
+        if (!currentRoutePath) {
```
neodyymi: This sould probably never happend if we do this right

Divinare:
Making one more change, then this should be ok

Divinare:
This can be reviewed now

src/components/sidebar/routePathView/NewRoutePathView.tsx
kronken: A wise man once said: Comments should be unnecessary

src/components/sidebar/routePathView/NewRoutePathView.tsx
kronken: I assume this is unnecessary if we move the initiation to the constructor.

src/components/sidebar/routePathView/NewRoutePathView.tsx
Divinare: Already fixed this

src/components/sidebar/routePathView/NewRoutePathView.tsx
Divinare: Already fixed this

src/components/sidebar/routePathView/NewRoutePathView.tsx
Divinare: Done at constructor

src/components/sidebar/routePathView/NewRoutePathView.tsx
Divinare: If I remove this if, then it says in render: "Object possibly undefined". So returning null if currentRoutePath is falsey should be fine

neodyymi: Does it need to be optional? I mean couldn't the ?-mark just be removed so it's always defined?

PR #397, opened 2018-11-21
305 log events to changelog

### LINKED ISSUE #305: Log a change to changelog
Adding an edit to the changelog

kronken:
Closes #305 

screenshot 2018-11-21 at 13 17 01

### DIFF
--- a/src/components/map/layers/new/NewRoutePathLayer.tsx
+++ b/src/components/map/layers/new/NewRoutePathLayer.tsx
@@ -21,16 +21,18 @@ interface IRoutePathLayerProps {
 @inject('routePathStore', 'geometryLogStore')
 @observer
 export default class RoutePathLayer extends Component<IRoutePathLayerProps> {
-    private renderLinks(links: IRoutePathLink[]|null, { isNeighbor }: any) {
-        if (!links) return;
+    private renderRoutePathLinks(
+        routePathLinks: IRoutePathLink[]|null,
+        { isNeighbor }: { isNeighbor: boolean }) {
+        if (!routePathLinks) return;
 
-        return links.map((routePathLink: IRoutePathLink, index) => {
+        return routePathLinks.map((routePathLink: IRoutePathLink, index) => {
             const nodeToRender = isNeighbor ? routePathLink.endNode : routePathLink.startNode;
             return (
             <div key={index}>
                 {this.renderNode(nodeToRender, routePathLink, { isNeighbor }, index)}
                 {/* Render last endNode of routePathLinks */}
-                { (index === links.length - 1 && !isNeighbor) &&
+                { (index === routePathLinks.length - 1 && !isNeighbor) &&
                     this.renderNode(routePathLink.endNode, routePathLink, { isNeighbor }, index)
                 }
                 {this.renderLink(routePathLink, { isNeighbor })}
@@ -39,12 +41,17 @@ export default class RoutePathLayer extends Component<IRoutePathLayerProps> {
         });
     }
 
-    private renderNode(node: INode, link: IRoutePathLink, { isNeighbor }: any, key: number) {
+    private renderNode
+        (
+            node: INode,
+            routePathLink: IRoutePathLink,
+            { isNeighbor } : { isNeighbor: boolean },
+            key: number) {
         const latLng = L.latLng(node.coordinates.lat, node.coordinates.lon);
         return (
             <Marker
                 key={`${key}-${node.id}`}
-                onClick={isNeighbor ? this.addLinkToRoutePath(link) : null}
+                onClick={isNeighbor ? this.addLinkToRoutePath(routePathLink) : null}
                 icon={this.getIcon({ isNeighbor }, node.type)}
                 position={latLng}
             />
@@ -82,35 +89,31 @@ export default class RoutePathLayer extends Component<IRoutePathLayerProps> {
         )}" />`;
     }
 
-    private renderLink(link: IRoutePathLink, { isNeighbor }: any) {
+    private renderLink(routePathLink: IRoutePathLink, { isNeighbor }: any) {
         const color = isNeighbor ? '#ca00f7' : '#00df0b';
         return (
             <Polyline
-                positions={link.positions}
-                key={link.id}
+                positions={routePathLink.positions}
+                key={routePathLink.id}
                 color={color}
                 weight={5}
                 opacity={0.8}
-                onClick={isNeighbor ? this.addLinkToRoutePath(link) : null}
+                onClick={isNeighbor ? this.addLinkToRoutePath(routePathLink) : null}
             />
         );
     }
 
-    private addLinkToRoutePath = (link: IRoutePathLink) => async () => {
-        const links = await this.queryNeighborLinks(link.endNode.id);
-        this.props.routePathStore!.setNeighborLinks(links);
+    private addLinkToRoutePath = (routePathLink: IRoutePathLink) => async () => {
+        const newRoutePathLinks =
+            await RoutePathLinkService.fetchLinksWithLinkStartNodeId(routePathLink.endNode.id);
+        this.props.routePathStore!.setNeighborRoutePathLinks(newRoutePathLinks);
         this.props.geometryLogStore!.pushToLog({
             action: logActions.ADD,
             entityName: entityNames.ROUTELINK,
-            newObject: link,
-            objectId: link.id,
+            newObject: routePathLink,
+            objectId: routePathLink.id,
         });
-        this.props.routePathStore!.addLink(link);
-    }
-
-    private async queryNeighborLinks(nodeId: string) {
-        const links = await RoutePathLinkService.fetchLinksWithLinkStartNodeId(nodeId);
-        return links;
+        this.props.routePathStore!.addLink(routePathLink);
     }
 
     private renderFirstNode() {
@@ -125,13 +128,13 @@ export default class RoutePathLayer extends Component<IRoutePathLayerProps> {
         if (!this.props.routePathStore!.routePath) return null;
         return (
             <>
-                {this.renderLinks(
+                {this.renderRoutePathLinks(
                     this.props.routePathStore!.routePath!.routePathLinks, { isNeighbor: false })}
                 {this.props.routePathStore!.routePath!.routePathLinks!.length === 0 &&
                     this.renderFirstNode()
                 }
                 {/* Neighbors should be drawn last */}
-                {this.renderLinks(
+                {this.renderRoutePathLinks(
                     this.props.routePathStore!.neighborLinks, { isNeighbor: true })};
             </>
         );
--- a/src/components/map/layers/new/NewRoutePathLayer.tsx
+++ b/src/components/map/layers/new/NewRoutePathLayer.tsx
@@ -7,14 +7,18 @@ import IRoutePathLink from '~/models/IRoutePathLink';
 import INode from '~/models/INode';
 import NodeType from '~/enums/nodeType';
 import { RoutePathStore } from '~/stores/routePathStore';
+import { GeometryLogStore } from '~/stores/geometryLogStore';
 import RoutePathLinkService from '~/services/routePathLinkService';
+import logActions from '~/enums/logActions';
+import entityNames from '~/enums/entityNames';
 import * as s from './newRoutePathLayer.scss';
 
 interface IRoutePathLayerProps {
     routePathStore?: RoutePathStore;
+    geometryLogStore?: GeometryLogStore;
 }
 
-@inject('routePathStore')
+@inject('routePathStore', 'geometryLogStore')
 @observer
 export default class RoutePathLayer extends Component<IRoutePathLayerProps> {
     private renderLinks(links: IRoutePathLink[]|null, { isNeighbor }: any) {
@@ -95,7 +99,12 @@ export default class RoutePathLayer extends Component<IRoutePathLayerProps> {
     private addLinkToRoutePath = (link: IRoutePathLink) => async () => {
         const links = await this.queryNeighborLinks(link.endNode.id);
         this.props.routePathStore!.setNeighborLinks(links);
-
+        this.props.geometryLogStore!.pushToLog({
+            action: logActions.ADD,
+            entityName: entityNames.ROUTELINK,
+            newObject: link,
+            objectId: link.id,
+        });
         this.props.routePathStore!.addLink(link);
     }
 
--- a/src/components/map/layers/new/NewRoutePathLayer.tsx
+++ b/src/components/map/layers/new/NewRoutePathLayer.tsx
@@ -41,12 +41,11 @@ export default class RoutePathLayer extends Component<IRoutePathLayerProps> {
         });
     }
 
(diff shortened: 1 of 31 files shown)


src/components/map/toolbar/EventLog.tsx
```diff
@@ -1,7 +1,12 @@
 import React from 'react';
 import { observer } from 'mobx-react';
-import { FaTimes, FaExclamation } from 'react-icons/fa';
+import classnames from 'classnames';
+import { FaTimes, FaExclamation, FaUndo } from 'react-icons/fa';
 import { FiClipboard } from 'react-icons/fi';
+import ILogEntry from '~/models/ILogEntry';
+import logActions from '~/enums/logActions';
```
Divinare: logTypes? Okay here action word fits better. But how about logActionType? At least that is not a verb.

src/components/map/toolbar/EventLog.tsx
```diff
@@ -14,16 +19,86 @@ interface IEventLogProps {
...
+        case logActions.ADD:
+            logActionClass = s.logActionAdd;
+            break;
+        case logActions.DELETE:
+            logActionClass = s.logActionDelete;
+            break;
+        case logActions.MOVE:
+            logActionClass = s.logActionMove;
+            break;
+        }
```
Divinare: There is no such events yet as DELETE / MOVE. Why not just log ADD events?

Imo, generally it is good to not code something that is not being used - or how can I test it, reviewing also a bit harder.

Also when the time comes when this is taken into use and I put a review comment on it, the coder says "it was already there"

src/components/map/toolbar/EventLog.tsx
```diff
@@ -14,16 +19,86 @@ interface IEventLogProps {
...
     constructor(props: IEventLogProps) {
         super(props);
         this.state = {
             isOpen: false,
             isAlertShown: false,
         };
+        this.scrollRef = React.createRef();
+    }
+
+    private getLogRow = (action: logActions | null, objectId: string, entity: entityNames) => {
```
Divinare: What is our practice for a method that returns JSX? I think it doesn't start with get

src/stores/geometryLogStore.ts
```diff
@@ -0,0 +1,41 @@
...
+import entityNames from '~/enums/entityNames';
+
+export class GeometryLogStore {
+    @observable private _log: ILogEntry[];
+
+    constructor() {
+        this._log = [];
+    }
+
+    @computed get log(): ILogEntry[] {
```
Divinare: get logs

src/factories/routePathLinkFactory.ts
```diff
@@ -33,7 +39,7 @@ class RoutePathLinkFactory {
             endNode: NodeFactory.createNode(link.solmuByLnkloppusolmu),
             positions: RoutePathLinkFactory.getPositions(link.geojson),
             isStartNodeTimeAlignmentStop: false,
-            id: 'new-12345',
+            id: RoutePathLinkFactory.getTemporaryRoutePathLinkId(),
```
Divinare: I think using increasing counter would be better. New-1, new-2, new-3, new-4, ...

src/components/map/toolbar/EventLog.tsx
```diff
@@ -14,16 +19,86 @@ interface IEventLogProps {
...
+        Object.entries(groupedById).forEach(([id, groupedLogs]: [string, ILogEntry[]]) => {
+            const first = groupedLogs[0];
+            res.push(this.getLogRow(
+                this.getSummaryAction(groupedLogs),
+                id,
+                first.entity,
+            ));
+        });
+
+        return res;
```
Divinare: This needs rewriting as it seems way too complex considering that we only show ADD events at this point.

Why not just use something like:

```
const events = this.props.GeometryLogStore!.events

return (
   { this.events.forEach.map(event) => {
        return (
              <EventLogRow event={event} />
        )
     }
   }
)
```

Have EventLogRow in its own component.

src/constants/index.ts
```diff
@@ -0,0 +1,5 @@
+const NEW_OBJECT_TAG = 'new';
+
+export {
+    NEW_OBJECT_TAG,
+};
```
Divinare: Why was this added? Remove.

At our code we can use new- prefix globally for all new objects that haven't been saved. If it needs documenting, cotributions.md is the right place.

src/stores/geometryLogStore.ts
Divinare: Or would "events" be better?

geometryLogStore -> geometryEventStore?

Divinare:
Review ^

Testing:

Change log is not scrollable. Scrolling on changeLog zooms the map in/out.

When creating a new routePath is closed, changeLog still shows those events of adding a routeLink. -> clear changeLog & minimize it?

src/components/map/layers/new/NewRoutePathLayer.tsx
```diff
@@ -7,14 +7,18 @@ import IRoutePathLink from '~/models/IRoutePathLink';
 import INode from '~/models/INode';
 import NodeType from '~/enums/nodeType';
 import { RoutePathStore } from '~/stores/routePathStore';
+import { GeometryLogStore } from '~/stores/geometryLogStore';
 import RoutePathLinkService from '~/services/routePathLinkService';
+import logActions from '~/enums/logActions';
+import entityNames from '~/enums/entityNames';
```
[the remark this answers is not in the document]
kronken: Why would we have these in separate directory when all other enums are in the same?

src/components/map/toolbar/EventLog.tsx
kronken: Action is both a noun and a verb, and I think in this context it is clear that the word is a noun.

src/components/map/toolbar/EventLog.tsx
```diff
@@ -1,7 +1,12 @@
 import React from 'react';
 import { observer } from 'mobx-react';
-import { FaTimes, FaExclamation } from 'react-icons/fa';
+import classnames from 'classnames';
+import { FaTimes, FaExclamation, FaUndo } from 'react-icons/fa';
 import { FiClipboard } from 'react-icons/fi';
+import ILogEntry from '~/models/ILogEntry';
+import logActions from '~/enums/logActions';
+import entityNames from '~/enums/entityNames';
```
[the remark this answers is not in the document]
kronken: Because the enum is a discription of the name of the entity, such as route, routePath or likewise. EntityName is a common term used to declare these things. The term has almost 1 million hits on google, compared to 400.000 for terms such as "entityType". https://www.google.fi/search?ei=iKX7W8zkCMOTsAGX5JjoDg&q=%22entityname%22&oq=%22entityname%22

src/components/map/toolbar/EventLog.tsx
kronken: I don't find anything suggesting that we have a practise for this in our contributions file: 
https://github.com/HSLdevcom/jore-map-ui/blob/master/Contributions.md

src/stores/geometryLogStore.ts
kronken: We have one log, multiple log entries. We don't have multiple logs, that would mean we have multiple different logs to store different things. Log vs Event is debatable, event could in this context indeed work better.

src/factories/routePathLinkFactory.ts
kronken: Sure, this I agree on.

src/components/map/toolbar/EventLog.tsx
kronken: Your suggestion would not take into consideration modifying the same item twice. Your suggestion would in that case print both events, when we have agreed on only showing the entry once.
Having EventLogRow in a separate component is a good advice.

src/constants/index.ts
kronken: I don't want us to have magic strings that we use for matching, that is only declared in contributions.md. We need to match this prefix to know which ones are added and which ones are old, and we can't start writing hard-coded prefixes.

src/enums/logActions.ts
```diff
@@ -0,0 +1,7 @@
+enum logActions {
+    ADD= 'Lisäys',
+    DELETE= 'Poistaus',
+    MOVE= 'Siirräys',
```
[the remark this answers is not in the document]
kronken: thanks

src/components/map/toolbar/EventLog.tsx
kronken: Okay, this is a bit two-sided. I understand your point, especially in the reviewing part. My view is that it is natural to add the different alternatives when coding a thing like this, in order to see how it would work in the end, and since it is the most effective way, since the coder has insight in the component that he is coding. I can leave the code there, but having it commented out. Which means it will need reviewing when uncommenting it.

src/stores/geometryLogStore.ts
kronken: events

src/components/map/layers/new/NewRoutePathLayer.tsx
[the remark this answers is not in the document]
kronken: entityNames are also used in separate functionalities, and we don't create new folder just for logActions

kronken:
Found bug that logging stops working after reinitialised routepath, created bug ticket: https://github.com/HSLdevcom/jore-map-ui/issues/430

src/components/map/layers/new/NewRoutePathLayer.tsx
```diff
@@ -102,6 +106,13 @@ export default class RoutePathLayer extends Component<IRoutePathLayerProps> {
         const newRoutePathLinks =
             await RoutePathLinkService.fetchLinksWithLinkStartNodeId(routePathLink.endNode.id);
         this.props.routePathStore!.setNeighborRoutePathLinks(newRoutePathLinks);
+        this.props.geometryLogStore!.pushToLog({
+            action: logActions.ADD,
+            entityName: entityNames.ROUTELINK,
+            newObject: routePathLink,
+            objectId: routePathLink.id,
+        });
```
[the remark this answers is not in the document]
kronken: Used this advice and created separate store for this, where everything is handled isolated from other things

src/components/map/toolbar/EventLog.tsx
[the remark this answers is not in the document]
Divinare: "routePath", "route", etc refer to the Models in Jore. They are all models whose type/name is IRoutePath, IRoute. So maybe both type/name can be used...

src/components/map/toolbar/EventLog.tsx
Divinare: I think we are still mixing two concepts, eventLog and showing what changes user has made (changeLog). EventLog !== changeLog. We are implementing eventLog here.

If we have the following case: user adds a link, removes a link, adds the link back again

EventLog shows:
Add link
Remove link
Add link

ChangeLog shows:
Add link

I made a more simplified code that you should use here since we are simply rendering through a list here:

```
    private renderEventRow(action: logActions, entity: entityNames) {
        let eventTypeClass = '';
        switch (action) {
        case logActions.ADD:
            eventTypeClass = s.eventTypeAdd;
            break;
        // Uncomment when we have use case for edit and deletion
        //
        // case logActions.DELETE:
        //     eventTypeClass = s.eventTypeDelete;
        //     break;
        // case logActions.MOVE:
        //     eventTypeClass = s.eventTypeMove;
        //     break;
        }

        return (
            <div className={s.event}>
                <div className={classnames(s.eventType, eventTypeClass)}>
                    {action}
                </div>
                <div className={s.eventContent}>
                    {entity}
                </div>
                <div className={s.eventButtons}>
                    <FaUndo />
                </div>
            </div>
        );
    }

    private renderEventList = () => {
        const events = GeometryEventStore!.events;
        return events.map(event: IEvent) => {
            this.renderEventRow(event.action, event.entity);
        }
    }
```

src/stores/geometryEventStore.ts
```diff
@@ -0,0 +1,103 @@
...
+                        action: eventTypes.ADD,
+                        entityName: entityNames.ROUTELINK,
+                        newObject: routePathLink,
+                        objectId: routePathLink.id,
+                    });
+                });
+        }
+    }
+
+    private initClearEventListOnPageChange() {
```
Divinare: Hmm, how about having to call this manually in NewRoutePath's componentWillUnmount and rename it as removeEvents()? I wouldn't make an asumption that we would always want to reset eventStore when sidebar page changes.

kronken: We have the possibility to modify the clear calling later, if we find the need for this. But I would suggest keeping it isolated as long as we dont see the need for extracting it.

src/components/map/toolbar/EventLog.tsx
```diff
@@ -1,7 +1,12 @@
 import React from 'react';
 import { observer } from 'mobx-react';
-import { FaTimes, FaExclamation } from 'react-icons/fa';
+import classnames from 'classnames';
+import { FaTimes, FaExclamation, FaUndo } from 'react-icons/fa';
 import { FiClipboard } from 'react-icons/fi';
+import ILogEntry from '~/models/IEvent';
+import logActions from '~/enums/eventTypes';
```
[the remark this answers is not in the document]
kronken: fixed

src/components/map/toolbar/EventLog.tsx
kronken: Okay, I did now remove the grouping logic, so we now have a simpler event log.

src/stores/geometryEventStore.ts
```diff
@@ -0,0 +1,115 @@
...
+        observe(
+            Navigator!.getStore(),
+            () => {
+                this.clearEvents();
+            },
+        );
+    }
+
+    private initRoutePathLinkObservable() {
+        let reactor : Lambda | null = null;
```
Divinare: routePathReactor

src/stores/geometryEventStore.ts
```diff
@@ -0,0 +1,115 @@
...
+        // Creating watcher which will trigger when RoutePathStore is initialized
+        // We cannot watch RoutePathStore!.routePath!.routePathLinks!
+        // before we know that it is defined
+        observe(
+            RoutePathStore!,
+            (change) => {
+                // We need to watch routePath, however, this object is created and
```
Divinare: if change.name === '_routePath return;

src/stores/geometryEventStore.ts
```diff
@@ -0,0 +1,103 @@
...
+          objectId: string,
+          oldObject?: object,
+          newObject?: object,
+      },
+      ) {
+        this._events.push({
+            action,
+            objectId,
+            postObject: newObject,
+            preObject: oldObject,
```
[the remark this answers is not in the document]
kronken: commented this

src/stores/geometryEventStore.ts
kronken: fiexd

src/stores/geometryEventStore.ts
kronken: fixed

Divinare:
Testing ok

PR #418, opened 2018-11-27
303 removed references to routepath geometry

neodyymi:
Backend changes required change to frontend.
* Removed routepath geometry from backend


Divinare:
Review ^

Issue #405, opened 2018-11-26
Network filter view to map


Divinare:
* Remove network filtering from /routes/ and add a button to (bottom left?) that opens filter network window just like mapLayers button

PR #421, opened 2018-11-27
405 network filter moved to mapControls

ahjyrkia:
closes https://github.com/HSLdevcom/jore-map-ui/issues/405

### DIFF
--- a/src/components/map/mapControls/MapLayersControl.tsx
+++ b/src/components/map/mapControls/MapLayersControl.tsx
@@ -40,16 +40,16 @@ export default class MapLayersControl extends React.Component
         NetworkStore.toggleTransitType(type);
     }
 
-    public toggleIsLinksVisible = () => {
-        NetworkStore.toggleIsLinksVisible();
+    public toggleLinkVisibility = () => {
+        NetworkStore.toggleLinkVisibility();
     }
 
-    public toggleIsNodesVisible = () => {
-        NetworkStore.toggleIsNodesVisible();
+    public toggleNodeVisibility = () => {
+        NetworkStore.toggleNodeVisibility();
     }
 
-    public toggleIsPointsVisible = () => {
-        NetworkStore.toggleIsPointsVisible();
+    public togglePointVisibility = () => {
+        NetworkStore.togglePointVisibility();
     }
 
     render() {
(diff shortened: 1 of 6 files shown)


src/components/map/mapControls/MapLayersControl.tsx
```diff
@@ -33,28 +36,78 @@ export default class MapLayersControl extends React.Component
         });
     }
 
+    public toggleTransitType = (type: TransitType) => {
+        NetworkStore.toggleTransitType(type);
+    }
+
+    public toggleIsLinksVisible = () => {
```
neodyymi: These names feel like bad english 😅 
Could they just be `toggleLinksVisibility` etc

neodyymi: I'm also not sure why these are wrapped in local functions when the ones from networkstore are selfexplanatory already..

src/components/map/mapControls/MapLayersControl.tsx
```diff
@@ -33,28 +36,78 @@ export default class MapLayersControl extends React.Component
...
-                        onClick={this.toggleRadioButton(option.SATELLITE)}
-                        checked={this.state.selectedOption === option.SATELLITE}
-                        text={option.SATELLITE}
-                    />
-                    <RadioButton
-                        onClick={this.toggleRadioButton(option.TERRAIN)}
-                        checked={this.state.selectedOption === option.TERRAIN}
-                        text={option.TERRAIN}
-                    />
+                    <div className={s.flexColumn}>
```
neodyymi: This div seems to be redundant.

src/components/map/mapControls/mapLayersControl.scss
```diff
@@ -24,6 +25,31 @@
         border: 1px solid $darkGrey;
         z-index: 1;
         padding: 10px;
+
```
neodyymi: There's a non-breaking space after `.mapLayersContainer`. Seems like it's been there before, but I think it should be removed asap.

ahjyrkia: Tried removing it. Didn't work. Added it back.

neodyymi:
Add the weird behaviour of the hover on some browsers as an issue and this is fine

PR #426, opened 2018-11-28
382 reusable graphql query functions

Divinare:
Now graphql functions are easilly maintanable. If for example a query field is added or removed to nodes all alku- & loppusolmu queries

The only drawback could be that for some cases you don't always have to query all of the information (but luckily apollo's cache can store that extra info and it can be used later)
   * but if we need a minimal query in the future, we can add a param to for example "getRouteQuery(queryHierarchies: number} that can limit how deep the query goes.

Maybe graphql.ts could have a better name? I didn't come up with better one though.


src/services/graphql.ts
```diff
@@ -0,0 +1,199 @@
+import gql from 'graphql-tag';
+
+// tslint:disable:max-line-length
+
+function getLineQuery() {
+    return (
+        gql`query getLineByLintunnus ($lineId: String!) {
+            linjaByLintunnus(lintunnus:$lineId) {
+                ${lineQueryFields}
+                reittisByLintunnus(first: 1, orderBy: REIVIIMPVM_DESC) {
```
kronken: Why limit the number of reittis?

src/services/graphql.ts
```diff
@@ -0,0 +1,199 @@
...
+}
+
+function getLinksQuery() {
+    return (
+        gql`query getNodesWithRoutePathLinkStartNodeId($nodeId: String!) {
+            solmuBySoltunnus(soltunnus: $nodeId) {
+                linkkisByLnkalkusolmu {
+                    nodes {
+                        ${routeLinkQueryFields}
+                        geojson
```
kronken: geojson is both declared here and in routeLinkQueryFields varible

src/services/graphql.ts
kronken: I know we probably only use the first. But I see this as a problem since we are trying to create generic functions. This is a function for getting a line. But, this query will only return one route. What if someone starts to use this function, and don't notice that we only take the first.
Maybe we could add a comment, just for notifying why we only take the first?

kronken: Or just modify it returning all routes

src/services/graphql.ts
```diff
@@ -0,0 +1,199 @@
...
+
+function getRoutePathQuery() {
+    return (
+        gql`query getRoutePath($routeId: String!, $startDate: Datetime!, $direction: String!) {
+            routePath: reitinsuuntaByReitunnusAndSuuvoimastAndSuusuunta(reitunnus: $routeId, suuvoimast: $startDate, suusuunta: $direction) {
+                ${routePathQueryFields}
+                reittiByReitunnus {
+                    lintunnus
+                }
+                reitinlinkkisByReitunnusAndSuuvoimastAndSuusuunta {
```
kronken: This is exactly identical to the code used on row 48, in the function above. Could we reuse this code?

neodyymi:
I'd rather see the module called `GraphqlQueries` Or something that wouldn't seem so much like an external node module to utilize graphql

src/services/graphql.ts
Divinare: Lets try to remove first: 1

src/services/graphql.ts
Divinare: const reitinlinkkiQuery = ...

src/services/graphql.ts
Divinare: removed this

Divinare:
Review fixes done.

PR #428, opened 2018-11-28
Initial edit network mode done

Divinare:
New view: NetworkView
* Content of that sidebar in initial will change in the future

* Renamed networkStore's setIsNodesVisible -> setNodeVisibility

### DIFF
--- a/src/components/map/toolbar/toolbarModeButtons.tsx
+++ b/src/components/map/toolbar/toolbarModeButtons.tsx
@@ -19,16 +19,16 @@ export default class ToolbarModeButtons extends React.Component<IToolbarModeButt
     }
 
     private initStores() {
-        if (navigator.getPathName() === subSites.network) {
-            toolbarStore.setEditMode(EditMode.NETWORK);
-        }
+        const editMode = navigator.getPathName() === subSites.network
+            ? EditMode.NETWORK : EditMode.LINE;
+        toolbarStore.setEditMode(editMode);
     }
 
     private selectLineMode() {
         toolbarStore.setEditMode(EditMode.LINE);
         const homeViewLink = routeBuilder
             .to(subSites.home)
-            .removeQueryParams()
+            .clear()
             .toLink();
         navigator.goTo(homeViewLink);
     }
@@ -37,7 +37,7 @@ export default class ToolbarModeButtons extends React.Component<IToolbarModeButt
         toolbarStore.setEditMode(EditMode.NETWORK);
         const networkViewLink = routeBuilder
             .to(subSites.network)
-            .removeQueryParams()
+            .clear()
             .toLink();
         navigator.goTo(networkViewLink);
     }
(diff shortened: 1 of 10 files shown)


src/components/map/toolbar/toolbarModeButtons.tsx
```diff
@@ -2,12 +2,41 @@ import React from 'react';
...
+import routeBuilder from '~/routing/routeBuilder';
+import subSites from '~/routing/subSites';
+import navigator from '~/routing/navigator';
 import { RadioButton } from '../../controls';
 import * as s from './toolbarModeButtons.scss';
 
 export default class ToolbarModeButtons extends React.Component {
-    private toggleSelectedMode = (option: EditMode) => () => {
-        toolbarStore.setEditMode(option);
+    constructor(props: any) {
```
kronken: Don't use props: any. Create dummy prop interface if we don't set any props manually, that is the way we have done in other parts of the project

src/routing/routeBuilderContext.ts
```diff
@@ -57,6 +57,11 @@ export default class RouteBuilderContext {
         return this;
     }
 
+    removeQueryParams() {
```
kronken: We already have this function, 10 rows below we have function clear() that does this exact thing. The functions of this class that modifies the params have the names: append(), remove(), set() and clear(). The name removeQueryParams does not fit in.
The current names are maybe not clear enough, and we could probably change them to something like: addQueryParam(), removeQueryParam(), setQueryParam(), clearQueryParams()

src/components/sidebar/networkView/NetworkView.tsx
```diff
@@ -0,0 +1,36 @@
...
+import * as s from './networkView.scss';
+
+interface INetworkViewProps {
+    networkStore?: NetworkStore;
+}
+
+@inject('networkStore')
+@observer
+class NetworkView extends React.Component<INetworkViewProps> {
+    constructor(props: any) {
```
kronken: Do not use props: any

src/components/sidebar/networkView/NetworkView.tsx
```diff
@@ -0,0 +1,36 @@
+
```
kronken: Remove empty row

src/components/map/toolbar/toolbarModeButtons.tsx
```diff
@@ -2,12 +2,41 @@ import React from 'react';
...
+        super(props);
+
+        this.initStores();
+    }
+
+    private initStores() {
+        if (navigator.getPathName() === subSites.network) {
+            toolbarStore.setEditMode(EditMode.NETWORK);
+        }
```
kronken: We could move the initial set of EditMode into this function, and leave the initial EditMode as undefined. Otherwise, with this change, we have two locations where we initially set the mode. Both in this function and in the store.

src/routing/routeBuilderContext.ts
Divinare: I know removeQueryParams doesn't fit in but I didn't want to do refactoring since I would have to change everything... Should've added TODO comments.

Divinare: Using clear() - didn't see it back then...

src/components/map/toolbar/toolbarModeButtons.tsx
Divinare: Removed init from store. Didn't want to initialize it as undefined (would have to add type EditMode|undefined) and undefined is not a valid value for it

Issue #434, opened 2018-12-03
Marker to its own component (NodeMarker.tsx)


Divinare:
Create a reusable nodeMarker component

props: position (latLng), color (optional), onContextMenu (optional), isDraggable (optional), radius (optional), nodeType (optional)

use at routePathLinkLayer, nodeLayer, newRoutePathLayer

kronken:
Hmm, maybe, in which way would this be different from the react-leaflet marker component we are using already?

kronken:
The react leaflet marker contains parameters such as position, onContextMenu, draggable...

PR #435, opened 2018-12-03
403 new route path start marker

### LINKED ISSUE #403: Create new routePath - add start marker of the route

Divinare:
Closes #403 

I moved MarkerLayer code into routePathLinkLayer like we have discused before

I think drawing marker should be put into its own Marker.tsx component. Card for creating Marker.tsx: #434

What do you think about #434?


src/components/map/layers/new/newRoutePathLayer.scss
```diff
@@ -35,6 +35,11 @@
 
 }
 
+.startMarker {
+    display: block;
+    opacity: 0.8;
+}
+
```
kronken: Remove extra empty lines

Issue #437, opened 2018-12-03
Network node filtering (frontend)


neodyymi:
Frontend part of #417

PR #438, opened 2018-12-03
nodes obey transit type selection

### LINKED ISSUE #417: Network node filtering
Filtering network routeLinks & linkPoints done in here: #375

---

Is this even possible? Can nodes have .lnkverkko property (transitType). If yes, it should be added to node property (backend task) and taken in use at frontend

---

Also should transitType be added as routePaths property? Maybe not, because we have it at line already and its typical in databases that parents share data with all of their children

neodyymi:
Closes #417 #437 

Backend part: https://gitlab.hsl.fi/jore/jore-map-backend/merge_requests/17

Might have trouble getting cache to refresh.. Try rebuilding/removing cache from geoserver and disable cache in browser developer tools.


src/components/map/layers/NetworkLayers.tsx
```diff
@@ -100,7 +101,16 @@ export default class NetworkLayers extends Component<INetworkLayersProps> {
...
-            solmu: (properties: any, zoom: any) => {
+            solmu: (properties: any, zoom: number) => {
+
+                const selectedTransitTypes = toJS(this.props.networkStore!.selectedTransitTypes);
+                if (properties.transittypes) {
+                    const transitTypes = TransitTypeHelper
+                        .convertTransitTypeCodesToTransitTypes(properties.transittypes.split(','));
+                    if (selectedTransitTypes.every(type => !transitTypes.includes(type))) {
+                        return this.getEmptyStyle();
+                    }
```
kronken: Hmm, would this be more understandable using:
if (!transitTypes.some(type => selectedTransitTypes.includes(type))) {
    return ...
}
Or, which do you think is easier to read?
Algorithmically speaking, my suggestion would be faster, since we don't need to loop through all items. But these arrays are so small, so it wont matter, use the one you find is easier to read.

src/components/map/layers/NetworkLayers.tsx
```diff
@@ -100,7 +101,16 @@ export default class NetworkLayers extends Component<INetworkLayersProps> {
 
         return {
             // Layer name 'solmu' is directly mirrored from Jore through geoserver
-            solmu: (properties: any, zoom: any) => {
+            solmu: (properties: any, zoom: number) => {
+
+                const selectedTransitTypes = toJS(this.props.networkStore!.selectedTransitTypes);
+                if (properties.transittypes) {
+                    const transitTypes = TransitTypeHelper
+                        .convertTransitTypeCodesToTransitTypes(properties.transittypes.split(','));
```
kronken: I suggested using array instead of string in your BE task, let's see what we agree on there before merging this parsing.

src/components/map/layers/NetworkLayers.tsx
neodyymi: This is a good point. I'll change it to some.

src/components/map/layers/NetworkLayers.tsx
neodyymi: Geoserver didn't seem to be able to handle an array from postgres. That was my intention at first, but I ended up needing to make it a strïng instead.

src/components/map/layers/NetworkLayers.tsx
```diff
@@ -100,7 +101,16 @@ export default class NetworkLayers extends Component<INetworkLayersProps> {
 
         return {
             // Layer name 'solmu' is directly mirrored from Jore through geoserver
-            solmu: (properties: any, zoom: any) => {
+            solmu: (properties: any, zoom: number) => {
+
+                const selectedTransitTypes = toJS(this.props.networkStore!.selectedTransitTypes);
+                if (properties.transittypes) {
+                    const transitTypes = TransitTypeHelper
```
Divinare: We already have "selectedTransitTypes so this should be names as "propertyTransitTypes" or something

src/util/transitTypeHelper.ts
```diff
@@ -17,4 +17,7 @@ export default class TransitTypeHelper {
             return TransitType.NOT_FOUND;
         }
     }
+    public static convertTransitTypeCodesToTransitTypes = (types: string[]) => {
+        return types.map(t => TransitTypeHelper.convertTransitTypeCodeToTransitType(t));
+    }
```
Divinare: Hmm, I'm not sure if it is worth creating an extra helper for just looping through array.

I would use:

const transitTypes = properties.transittypes.split(',').map(TransitTypeHelper.convertTransitTypeCodeToTransitType);

and remove this helper.

src/components/map/layers/NetworkLayers.tsx
```diff
@@ -100,7 +101,16 @@ export default class NetworkLayers extends Component<INetworkLayersProps> {
 
         return {
             // Layer name 'solmu' is directly mirrored from Jore through geoserver
-            solmu: (properties: any, zoom: any) => {
+            solmu: (properties: any, zoom: number) => {
+
+                const selectedTransitTypes = toJS(this.props.networkStore!.selectedTransitTypes);
+                if (properties.transittypes) {
```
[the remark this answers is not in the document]
neodyymi: It's reflected from geoserver and postgres. Could make it transit_types if that would be better.

src/components/map/layers/NetworkLayers.tsx
neodyymi: Good point, renaming

src/util/transitTypeHelper.ts
neodyymi: I sort of disagree. Feels cleaner having this as a static helper. It has a single purpose and we will probably end up needing it elsewhere too.

neodyymi:
We have to figure out how this caching works.. It's sort of a pain in these situations..

src/util/transitTypeHelper.ts
Divinare: For me using .map is easier to distinct than "s" letter inside function convertTransitTypeCode_s_ToTransitType

But this isn't a big deal so change it or leave it like it is.

src/components/map/layers/NetworkLayers.tsx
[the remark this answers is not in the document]
Divinare: Ah I see. Also you could say
const propertyTransitTypes = properties.transittypes;
if (propertyTransitTypes) ...

but up to you, change or not

Divinare:
Testing ok

PR #440, opened 2018-12-04
434 node marker to its own component

### LINKED ISSUE #434: Marker to its own component (NodeMarker.tsx)

Divinare:
Closes #434

Now creating marker & styling it is done in a single place
   * cleaner code in components using ```<Marker>```
   * guarantees different markers to have the same styles everywhere.


Contributions.md
```diff
@@ -22,7 +22,7 @@
 * Use classnames if there are multiple classes
 * Use async await instead of promises
 * Don't use .bind in jsx, use arrow functions instead
-
+* Use render prefix in methods that return JSX or HTML
```
neodyymi: What does this mean?

src/components/map/layers/NodeMarker.tsx
```diff
@@ -0,0 +1,152 @@
...
+        return (
+            <Marker
+                zIndexOffset={VERY_HIGH_Z_INDEX}
+                icon={icon}
+                position={this.props.latLng}
+            />
+        );
+    }
+
+    private renderNodeStopCircle() {
```
neodyymi: Maybe radius or something more describing instead of circle?

src/components/map/layers/NodeMarker.tsx
```diff
@@ -0,0 +1,152 @@
...
+            <Circle
+                className={s.stopCircle}
+                center={this.props.latLng}
+                radius={radius}
+            />
+        );
+    }
+
+    render() {
+        const nodeType = this.props.nodeType;
```
neodyymi: This is only used in the following if statement. I think this is a waste of a row to make a const of it.

src/components/map/layers/NodeMarker.tsx
neodyymi: Meaning: I know it's a circle and all, but it's only used to display the radius of the node.

src/components/map/layers/NodeMarker.tsx
```diff
@@ -0,0 +1,152 @@
...
+    private renderMarker(html: any) {
+        const divIconOptions : L.DivIconOptions = {
+            html,
+            className: s.node,
+        };
+
+        return new L.DivIcon(divIconOptions);
+    }
+
+    private renderStartMarker() {
```
neodyymi: What if this took color and latlng as parameters instead. Would be cleaner in my opinion when called with those values from render.

src/components/map/layers/NodeMarker.tsx
```diff
@@ -0,0 +1,152 @@
...
+
+@inject('mapStore')
+@observer
+export default class NodeMarker extends Component<INodeMarkerProps> {
+    static defaultProps = {
+        isSelected: false,
+        isDraggable: false,
+    };
+
+    private renderMarkerLabel = () => {
```
neodyymi: Rather have it take stop and zoom as parameters. Would be cleaner in my opinion when called with those values from render.

src/components/map/layers/new/NewRoutePathLayer.tsx
```diff
@@ -50,47 +44,18 @@ export default class RoutePathLayer extends Component<IRoutePathLayerProps> {
...
-            }
-        }
-        return `<div class="${classnames(
-            s.nodeBase,
-            className,
-        )}" />`;
-    }
-
     private renderLink(routePathLink: IRoutePathLink, { isNeighbor }: any) {
         const color = isNeighbor ? '#ca00f7' : '#00df0b';
```
neodyymi: Could remove these magic numbers while you're at it

src/enums/nodeType.ts
```diff
@@ -1,10 +1,12 @@
 enum NodeType {
+    START = 'S',
     STOP = 'P',
     CROSSROAD = 'X',
     MUNICIPALITY_BORDER = '-',
     INVALID = 'invalid',
     DISABLED = 'E',
     TIME_ALIGNMENT = '1',
+    IS_NEIGHBOR = 'N',
```
neodyymi: What does this mean?

Divinare:
Review fixes done

src/enums/nodeType.ts
Divinare: Means node is type can be neighbor. At the moment neighbor type is used when creating new routePaths, purple large nodes (neighbor nodes) have nodeType='N'

Divinare: This could be renamed as "IS_NEW_ROUTEPATH_NEIGHBOR" but i think the property name should not describe where it is being used

neodyymi: Hmm okay, I was just a bit confused since a lot of this refers to how jore works. I guess this makes sense.

PR #442, opened 2018-12-04
424 form validation

### LINKED ISSUE #424: RPV - 1 Solution for enabling validation of forms, with one simple use case
The input field could have a hint saying the restrictions, forcing the input to be compliant or highlighting with red if the input is invalid.

Required to be implemented using a generic approach, probably using a higher order input field.

kronken:
Closes #424 
Test by opening routepath, and modify nameFI and length fields.
Empty fields should throw error,
Too long fields should throw error,
Text in length field should throw error,
Any error should disable the save button


src/validation/FormValidator.ts
```diff
@@ -0,0 +1,29 @@
...
+                value,
+            },
+            {
+                value: rule,
+            },
+            ruleTranslations,
+            );
+
+        const isValid = !!validator.passes();
+        const firstErrorMessage = validator.errors.first('value');
```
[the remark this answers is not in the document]
kronken: The UI would just be so much more complicated. And I think usually in similar UI:s there is only one error per input.

src/validation/FormValidator.ts
```diff
@@ -0,0 +1,29 @@
+import Validator from 'validatorjs';
+import ruleTranslations from './validationRuleTranslations';
+
+export interface IValidationResult {
+    isValid: boolean;
```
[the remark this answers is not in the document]
kronken: Yeah, you are correct. The reason why I did it this way, is since validatorjs official function for validating is .passes(). and therefore I would like to use this function. By only using error message, we are not using their preferred way to validate, and I'm not certain about the results in that case. I don't for example know, if the validator can invalidate a field, but not be able to give an error message.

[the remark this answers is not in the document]
kronken: So, I would like to keep it this way

src/components/sidebar/routePathView/RoutePathView.tsx
```diff
@@ -24,6 +25,7 @@ interface IRoutePathViewState {
     hasModifications: boolean;
     routePath: IRoutePath | null;
     isLoading: boolean;
+    invalidFieldsDictionary: object;
```
ahjyrkia: I think it is more common just to use "map" rather than "dictionary". If that is what the word dictionary is implying here.

kronken: Yes, you are correct, map is probably a better name

kronken: This is now fixed

src/components/sidebar/routePathView/RoutePathView.tsx
```diff
@@ -83,11 +86,28 @@ class RoutePathView extends React.Component<IRoutePathViewProps, IRoutePathViewS
         this.setState({ isLoading: false });
     }
 
-    public onChange = (property: string, value: any) => {
+    private markFieldIsValid = (field: string, isValid: boolean) => {
```
ahjyrkia: I think this needs renaming. Maybe refer to the state change it does so something like markInvalidFields or setInvalidFields?

PR #449, opened 2018-12-05
373 routepath information

neodyymi:
* All information in `RoutePathViewForm` is now fetched from backend
* `RoutePathView` is now available at _/routepath/routeid,starttime,direction/_


src/services/graphqlQueries.ts
```diff
@@ -129,6 +129,9 @@ const routePathQueryFields = `
     suupaapaik
     suukuka
     suupaapaikr
+    suupituus
+    suuhis
```
kronken: What are we using suuhis for?

neodyymi: Nothing. I checked and the only values are null or 0.. So not sure what this is for :)
Removing it.

Issue #457, opened 2018-12-10
EN - 2 part 2, Clicking node


Divinare:
Clicking a node

Highlight clicked node & network links that have either alku / loppusolmu of that node and also linkPoints. Highlighting means hiding the real node & links below highlighted node & links.

Open clicked node to the sidebar

This is the network's node, not the node that belongs to routePath -> still, use nodeView.tsx?
Fetch some data of that clicked node to the NodeView

EDIT: not using nodeView .tsx yet. Maybe this needs some UI planning first

Issue #456, opened 2018-12-10
ToolHelp for different tool phases


Divinare:
For example:
Phase 1: "Valitse solmu kartalta"

Phase 2: "Jatka reitin suunnan muodostamista valitsemalla seuraavia solmuja"

What if you could add toolHelp into the sidebar?

Also add a topic to toolHelp(?)

About implementation:

- How about adding a "renderToolHelpContent()" method to each tool? That would return an object that contains toolHelp texts according to the current tool phase

That method could return an object of:

{
  topic: "Uuden reitin suunnan luonti"
  description: "Valitse reitin suunnan ensimmäinen solmu kartalta"
}

And at toolHelp call that method.

EDIT: add BaseTool.toolphase: number? That could be used to display the current toolHelp. Also renderToolHelpContent could return the content according to the current phase.

Also at EditRoutePathLayerNode, dont display extendRpToolHighlight if toolphase is not 1.

onEscapePress, set tool phase as 1?

Issue #447, opened 2018-12-05
Refreshing network layers not working correctly


Divinare:
Blocked by #409

Helpful links:
https://github.com/Leaflet/Leaflet/issues/5441

https://github.com/Leaflet/Leaflet/issues/5442

One way to "resolve" this bug would be to set nodes always as interactive and at on networkClick methods check if there should be a click event or not (return if isNetworkNodesInteractive === false)

---

This might be because of leaflet.vectorgrid.

New routePath nodes not clickable bug:
1) at frontpage, put network nodes on
2) go to new rotuePath creation
3) nodes are not clickable

Nodes have pointer cursor when they shouldn't bug:
1) go to new routePath creation
2) network nodes are clickable
3) go back to frontpage
4) network nodes still have cursor (interactive={false} didn't update)

This might be because of our leaflet.vectorgrid is buggy. We currently have a custom leaflet.vectorgrid library in use - that is because click event to vectorLayers didn't work before

PR #458, opened 2018-12-10
409 edit network tool init

### LINKED ISSUE #409: EN - 2 part 1, Tool selecting
When network is activated
* Hide shown routePaths

When editNetworkNodeTool is activated
* Make nodes larger

Divinare:
Closes #409 

Continues at #457 

Also keep in mind this bug: #447 

Now the tool logic (click handling etc) can be written to different tools. BaseTool declares the interface that each tool will implement

Created a new card: #456 (so that each tool would know their own toolHelp texts and we would have a different toolHelp text according to the current toolHelp phase.)


src/stores/networkStore.ts
```diff
@@ -9,17 +9,24 @@ const TRANSIT_TYPES = [
...
+    normal,
+    large,
+}
+
 export class NetworkStore {
     @observable private _selectedTransitTypes: TransitType[];
     @observable private _isLinksVisible: boolean;
     @observable private _isNodesVisible: boolean;
     @observable private _isPointsVisible: boolean;
+    @observable private _nodeSize:NodeSize;
```
kronken: missing whitespace

src/stores/toolbarStore.ts
```diff
@@ -1,41 +1,77 @@
...
     @action
-    public toggleTool(tool: ToolbarTool) {
-        if (!this.isDisabled(tool)) {
-            this._activeTool = (this._activeTool === tool) ? ToolbarTool.None : tool;
+    public selectTool(tool: ToolbarTool | null) {
+        if (this._selectedTool) {
+            this._selectedTool.deactivate();
         }
+
+        // deselect current tool
```
kronken: unselect

src/stores/toolbarStore.ts
```diff
@@ -1,41 +1,77 @@
...
+        if (tool === null || (this._selectedTool && this._selectedTool.toolType === tool)) {
+            this._selectedTool = null;
+            return;
+        }
+        const foundTool = _.find(TOOLS, (_tool) => {
```
kronken: const foundTool = TOOLS.find(tool => tool.toolType === tool);
Do not use lodash if not necessary

src/components/map/layers/NetworkLayers.tsx
```diff
@@ -28,9 +27,11 @@ enum NodeColors {
 }
 
 interface INetworkLayersProps {
+    editNetworkStore?: EditNetworkStore;
```
kronken: Is editNetworkStore used?

src/components/map/layers/NetworkLayers.tsx
```diff
@@ -150,34 +153,20 @@ export default class NetworkLayers extends Component<INetworkLayersProps> {
...
-        return this.props.routePathStore!.isCreating
-            && hasRoutePathLinks;
-    }
-
     render() {
         const selectedTransitTypes = this.props.networkStore!.selectedTransitTypes;
+        const nodeSize = this.props.networkStore!.nodeSize;
+
+        const selectedTool = this.props.toolbarStore!.selectedTool;
+        let isNetworkNodesInteractive;
```
kronken: This is equal to having : any, please use typings

src/enums/toolbarTool.ts
```diff
@@ -1,9 +1,10 @@
+// TODO: rename as Tool (?)
```
kronken: This logic is so global, so I think tool would not be specific enough

src/stores/toolbarStore.ts
```diff
@@ -1,41 +1,77 @@
...
+const TOOLS = [
+    new AddNetworkNodeTool(),
+    new AddNewRoutePathTool(),
+    new CopyTool(),
+    new DivideLinkTool(),
+    new EditNetworkNodeTool(),
+    new PrintTool(),
+];
+
+// TODO: Rename ToolbarStore -> toolStore?
```
kronken: Hmm, I dont think toolStore is specific enough.

src/tools/AddNewRoutePathTool.ts
```diff
@@ -0,0 +1,42 @@
...
+/**
+ * Tool for creating new routePath
+ */
+export default class AddNewRoutePathTool implements BaseTool {
+    public toolType = ToolbarTool.AddNewRoutePath;
+    public activate() {}
+    public deactivate() {}
+
+    public isNetworkNodesInteractive() {
+        const hasRoutePathLinks =
```
kronken: I think the code style have been to have an indent on the two following rows

src/tools/AddNewRoutePathTool.ts
```diff
@@ -0,0 +1,42 @@
...
+        RoutePathStore!.routePath &&
+        RoutePathStore!.routePath!.routePathLinks!.length === 0;
+
+        return RoutePathStore!.isCreating && hasRoutePathLinks;
+    }
+
+    public onNetworkNodeClick = async (clickEvent: any) => {
```
kronken: Do not use any

src/tools/EditNetworkNodeTool.ts
```diff
@@ -0,0 +1,49 @@
...
+    public deactivate() {
+        EditNetworkStore.setIsWaitingForNodeSelection(false);
+        NetworkStore.setNodeSize(NodeSize.normal);
+    }
+
+    isNetworkNodesInteractive() {
+        return EditNetworkStore!.isWaitingForNodeSelection;
+    }
+
+    onNetworkNodeClick = async (clickEvent: any) => {
```
kronken: do not use any

src/tools/EditNetworkNodeTool.ts
```diff
@@ -0,0 +1,49 @@
...
+        // tslint:disable-next-line
+        console.log('at EditNetworkNodeTools onNetworkNodeClick');
+        // TODO: click handking.
```
kronken: handling

src/components/map/layers/NetworkLayers.tsx
```diff
@@ -150,34 +153,20 @@ export default class NetworkLayers extends Component<INetworkLayersProps> {
...
+        let onNetworkNodeClick;
+        if (selectedTool) {
+            isNetworkNodesInteractive = selectedTool.isNetworkNodesInteractive ?
+                selectedTool.isNetworkNodesInteractive() : false;
+            onNetworkNodeClick = selectedTool.onNetworkNodeClick ?
+            selectedTool.onNetworkNodeClick : void 0;
```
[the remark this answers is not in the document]
Divinare: Change to what? I can add a comment "void 0 = empty function"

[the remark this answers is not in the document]
kronken: Would:
() => {}
work?

src/stores/toolbarStore.ts
Divinare: https://english.stackexchange.com/questions/18465/unselect-or-deselect

Dictionaries (Merriam-Webster and New Oxford American Dictionary) have deselect but not unselect. The NOAD defines deselect as “turn off (a selected feature) on a list of options on a computer menu”, which is what you want.

Unselected, on the other hand, is used to qualify something that has not been selected, not something that was selected and isn't anymore.

src/tools/AddNewRoutePathTool.ts
Divinare: Couldn't find types for these click events (this is how they were before). I tried different types but couldnt get one working.

Found only this: https://github.com/mhasbie/react-leaflet-vectorgrid
(that isnt mentioning type...)

But I added type for this:
gridLayer.on('click', (event: L.LeafletEvent) => {

src/stores/toolbarStore.ts
kronken: interesting

src/tools/AddNewRoutePathTool.ts
kronken: ok

src/stores/toolbarStore.ts
kronken: I just see TODO:s as something that needs to be fixed. Renaming it to toolStore is not something that needs to be fixed, it is something that could be discussed. And if we have that discussion, I would probably suggest MapTools as a name.

Divinare: If i add a question mark that usually means "this probably needs to be fixed"

But yes you are right, mapTools could be a better name than just tool. Maybe rename enums/toolbarTool -> enums/mapTool

But as this isn't 100% sure yet we can remove this comment

src/stores/editNetworkStore.ts
```diff
@@ -0,0 +1,23 @@
+import { action, computed, observable } from 'mobx';
+
+export class EditNetworkStore {
```
neodyymi: Is this necessary to have separate from NetworkStore?

src/stores/mapStore.ts
```diff
@@ -90,26 +81,15 @@ export class MapStore {
...
-    @action
-    public cycleCoordinateSystem() {
-        this._displayCoordinateSystem =
-            GeometryService.nextCoordinateSystem(this._displayCoordinateSystem);
-    }
-
     @action
-    setNodeSize(nodeSize: NodeSize) {
-        this._nodeSize = nodeSize;
+    public setDisplayCoordinateSystem(value: CoordinateSystem) {
```
neodyymi: If we're really choosing to ban setters, lets have it said in `contributions.md`

src/stores/toolbarStore.ts
```diff
@@ -1,41 +1,76 @@
...
+    new AddNewRoutePathTool(),
+    new CopyTool(),
+    new DivideLinkTool(),
+    new EditNetworkNodeTool(),
+    new PrintTool(),
+];
+
 export class ToolbarStore {
-    @observable private _activeTool: ToolbarTool;
+    @observable private _selectedTool: BaseTool|null;
```
neodyymi: `_selectedTool?` rather? And undefined instead of null.

src/stores/toolbarStore.ts
```diff
@@ -1,41 +1,76 @@
...
 
+import BaseTool from '~/tools/BaseTool';
+import AddNetworkNodeTool from '~/tools/AddNetworkNodeTool';
+import AddNewRoutePathTool from '~/tools/AddNewRoutePathTool';
+import CopyTool from '~/tools/CopyTool';
+import DivideLinkTool from '~/tools/DivideLinkTool';
+import EditNetworkNodeTool from '~/tools/EditNetworkNodeTool';
+import PrintTool from '~/tools/PrintTool';
+
+const TOOLS = [
```
neodyymi: Might be more simple to have object instead of array. That way there would be no need for `_.find`. Now it seems overly elaborate way of finding the tool specified.

src/stores/toolbarStore.ts
neodyymi: Having TOOLS as an object and selecting the tool by key would be even simpler.

src/stores/toolbarStore.ts
```diff
@@ -1,41 +1,76 @@
...
+            throw new Error('Tried to select tool that was not found');
+        }
+        this._selectedTool = foundTool;
+        this._selectedTool.activate();
     }
 
-    public isActive(tool: ToolbarTool): boolean {
-        return this._activeTool === tool;
+    public isSelected(tool: ToolbarTool): boolean {
+        if (!this._selectedTool) return false;
```
neodyymi: Is this necessary?

src/components/App.tsx
```diff
@@ -35,6 +40,16 @@ class App extends React.Component<IAppProps, IAppState> {
...
     }
 
+    public componentWillMount() {
+        this.initStores();
+    }
+
+    private initStores() {
+        const editMode = navigator.getPathName() === subSites.network
+            ? EditMode.NETWORK : EditMode.LINE;
+        this.props.toolbarStore!.setEditMode(editMode);
```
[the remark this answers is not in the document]
Divinare: You mean a computed value that would read from navigator.getPathName()? That might work.

src/components/map/layers/NetworkLayers.tsx
```diff
@@ -18,7 +16,7 @@ enum GeoserverLayer {
     Point = 'piste',
 }
 
-// TODO: use .scss for these?
+// TODO: import these from NodeMarker?
```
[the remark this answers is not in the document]
Divinare: Fixed comment as:
// TODO: import these from NodeMarker's .scss

src/components/map/layers/VectorGridLayer.tsx
```diff
@@ -38,7 +40,10 @@ class VectorGridLayer extends GridLayer<IVectorGridLayerProps> {
 
     updateLeafletElement(fromProps: IVectorGridLayerProps, toProps: IVectorGridLayerProps) {
         super.updateLeafletElement(fromProps, toProps);
-        if (!this.areArraysEqual(fromProps.selectedTransitTypes, toProps.selectedTransitTypes)) {
+        // TODO: consider passing a single value "shouldUpdate" and
+        // redraw layers according to that variable
```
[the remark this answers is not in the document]
Divinare: You can pass any props to vectorGridLayer you want. This comment was to move redraw check to NetworkLayers.tsx by using "shouldUpdate" value

src/components/map/toolbar/toolbarHelp.tsx
```diff
@@ -1,32 +1,37 @@
...
-    private getToolbarHelpContent(tool: ToolbarTool) {
-        const textContainers = {
-            edit: (
+        switch (selectedTool.toolType) {
+        case ToolbarTool.AddNewRoutePath:
+            return (
+                <div>
+                    Muodostaaksesi reitin suunnan, valitse kartalta aloitus-solmu.
+                    Tämän jälkeen jatka reitin suunnan muodostamista valitsemalla seuraavia solmuja.
+                </div>
```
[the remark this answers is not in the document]
Divinare: Yes! I created a card for this already (#456). Mentioned about this at PR description

src/components/map/toolbar/toolbarHelp.tsx
```diff
@@ -1,32 +1,37 @@
...
-            return textContainers.edit;
+            );
         default:
             return null;
         }
     }
 
     render() {
-        const toolbarHelpContent = this.getToolbarHelpContent(this.props.tool);
+        const toolbarHelpContent = this.renderToolbarHelpContent();
```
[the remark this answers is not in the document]
Divinare: I was thinking about that before... This needs more refactoring than that however because of class s.toolbarHelp should be around the <div>'s

But how those texts are brought here will be fixed in #456 so maybe I just add a TODO comment there

src/stores/toolbarStore.ts
Divinare: I think null is better choice here because it is a valid value, "no tool selected". For me undefined would refer to something not being initialized

src/components/App.tsx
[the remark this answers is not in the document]
neodyymi: Yeah that's what came into my mind when reading this

src/stores/toolbarStore.ts
Divinare: Fixing.

src/components/map/toolbar/toolbarHelp.tsx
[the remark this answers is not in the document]
neodyymi: My bad :D

src/stores/toolbarStore.ts
neodyymi: Wouldn't undefined also mean that the current tool is undefined as in not set?

src/stores/editNetworkStore.ts
Divinare: Hmm, i don't even think we need this store anymore. But in the future I was planning to use it. Removing it now

src/stores/toolbarStore.ts
Divinare: Maybe we should all generally agree on this how to use null/undefined.

Divinare:
Moved tools under /map/ now. I still disagree however because they are under /components/ now, tools are only referenced from toolbarStore. But we can change it later if we decide to

PR #460, opened 2018-12-10
Refactored IModels to use questionMark instead on null

### LINKED ISSUE #445: Refactor all IModels to use ? in optional properties instead of null
* Add a note to contributions.md
* With this way you don't have to say: const stop = startNode.stop ? startNode.stop : undefined;
you can just say stop={startNode.stop}

Divinare:
Small pr.

Closes #445

### DIFF
--- a/src/components/map/layers/new/NewRoutePathLayer.tsx
+++ b/src/components/map/layers/new/NewRoutePathLayer.tsx
@@ -20,8 +20,8 @@ interface IRoutePathLayerProps {
 @observer
 export default class RoutePathLayer extends Component<IRoutePathLayerProps> {
     private renderRoutePathLinks(
-        routePathLinks: IRoutePathLink[]|null,
-        { isNeighbor }: { isNeighbor: boolean }) {
+        { isNeighbor }: { isNeighbor: boolean },
+        routePathLinks?: IRoutePathLink[]) {
         if (!routePathLinks) return;
 
         return routePathLinks.map((routePathLink: IRoutePathLink, index) => {
@@ -51,7 +51,7 @@ export default class RoutePathLayer extends Component<IRoutePathLayerProps> {
             <NodeMarker
                 key={`${key}-${node.id}`}
                 nodeType={nodeType}
-                onClick={isNeighbor ? this.addLinkToRoutePath(routePathLink) : undefined}
+                onClick={isNeighbor ? this.addLinkToRoutePath(routePathLink) : () => {}}
                 latLng={latLng}
             />
         );
@@ -66,7 +66,7 @@ export default class RoutePathLayer extends Component<IRoutePathLayerProps> {
                 color={color}
                 weight={5}
                 opacity={0.8}
-                onClick={isNeighbor ? this.addLinkToRoutePath(routePathLink) : null}
+                onClick={isNeighbor ? this.addLinkToRoutePath(routePathLink) : () => {}}
             />
         );
     }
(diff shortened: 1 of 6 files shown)


src/components/map/layers/new/NewRoutePathLayer.tsx
```diff
@@ -20,8 +20,8 @@ interface IRoutePathLayerProps {
 @observer
 export default class RoutePathLayer extends Component<IRoutePathLayerProps> {
     private renderRoutePathLinks(
-        routePathLinks: IRoutePathLink[]|null,
-        { isNeighbor }: { isNeighbor: boolean }) {
+        { isNeighbor }: { isNeighbor: boolean },
```
neodyymi: Why is this an object?

src/services/routeService.ts
```diff
@@ -15,10 +15,9 @@ export interface IMultipleRoutesQueryResult {
 }
 
 export default class RouteService {
-    // TODO: refactor undefined to null?
-    public static async fetchRoute(routeId: string): Promise<IRoute | undefined> {
+    public static async fetchRoute(routeId: string): Promise<IRoute | null> {
```
neodyymi: well, this depends on which advice we follow on this undefined vs null thing..

src/components/map/layers/new/NewRoutePathLayer.tsx
neodyymi: If we want to explicitly explain what something is, rather do `const isNeighbor = true` when calling this

neodyymi: Ok, I'm finding references to doing this.. But they usually just take one object with everything instead of an object with one value.

src/services/routeService.ts
kronken: But, I think we this far have used null in similar cases. So, I would suggest using null here as well, just to keep it consistent. Then we might refactor all of those nulls to undefined if we later agree on that.

src/components/map/layers/new/NewRoutePathLayer.tsx
Divinare: Fixed, removed that ugly isNeighbor boolean completely.

src/components/map/layers/new/NewRoutePathLayer.tsx
```diff
@@ -80,11 +113,11 @@ export default class RoutePathLayer extends Component<IRoutePathLayerProps> {
     }
```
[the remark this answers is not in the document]
Divinare: Naming there is just fine:
```
const newRoutePathLinks =
            await RoutePathLinkService.fetchAndCreateRoutePathLinksWithStartNodeId(
                routePathLink.endNode.id);
        this.props.routePathStore!.setNeighborRoutePathLinks(newRoutePathLinks);
```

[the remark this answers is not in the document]
Divinare: "i.e. Why is stuff only used here being put through mobx store?"

Neighbor links are set from AddNetRoutePathTool:

RoutePathStore!.setNeighborRoutePathLinks(routePathLinks);

But that is indeed not part of this review.

src/components/map/layers/new/NewRoutePathLayer.tsx
```diff
@@ -19,54 +19,87 @@ interface IRoutePathLayerProps {
...
+            const nodeToRender = routePathLink.startNode;
             return (
             <div key={index}>
-                {this.renderNode(nodeToRender, routePathLink, { isNeighbor }, index)}
+                {this.renderNode(nodeToRender, index)}
                 {/* Render last endNode of routePathLinks */}
-                { (index === routePathLinks.length - 1 && !isNeighbor) &&
-                    this.renderNode(routePathLink.endNode, routePathLink, { isNeighbor }, index)
+                { (index === routePathLinks.length - 1) &&
+                    this.renderNode(routePathLink.endNode, index)
```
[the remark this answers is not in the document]
Divinare: The id is different, the other has index + startNode.id and the other index + endNode.id

PR #462, opened 2018-12-11
446 save new routepath

### LINKED ISSUE #446: Save new routePath
* Send from frontend to backend
* At our backend, save it to jore (is there a test jore that can be used?) and after save, call update functions to get that saved routePath to our backend

Also PR in BE: https://gitlab.hsl.fi/jore/jore-map-backend/merge_requests/18

kronken:
Closes #446 

BE: https://gitlab.hsl.fi/jore/jore-map-backend/merge_requests/18


src/components/sidebar/routePathView/NewRoutePathView.tsx
```diff
@@ -33,24 +37,22 @@ class NewRoutePathView extends React.Component<INewRoutePathViewProps, INewRoute
...
-            };
+        const oldRoutePath = this.props.routePathStore!.routePath;
+        let newRoutepath: IRoutePath;
+        if (!oldRoutePath) {
+            newRoutepath = this.createNewRoutePath();
         } else {
-            this.state = {
-                currentRoutePath,
-            };
+            newRoutepath = RoutePathFactory.createNewRoutePathFromOld(oldRoutePath);
```
Divinare: I think this wouldve been enough:
newRoutePath.startTime.setDate(newRoutePath.startTime.getDate() + 1);

But anyway that createNewRoutePathFromOld should be removed when we get timestamp validation working. I suggest to add this: // TODO: remove when timestamp validation works

src/services/routePathService.ts
```diff
@@ -37,4 +37,9 @@ export default class RoutePathService {
         const apiClient = new ApiClient();
         return await apiClient.updateObject(entityName.ROUTEPATH, routePath);
     }
+
+    public static async createRoutePath(routePath: IRoutePath) {
```
Divinare: Hmm, maybe I said about this before, but I would rename as:

update existing item:
updateRoutePath & apliClient.updateRoutePath

add new item:
saveRoutePath & apiClient.saveRoutePath

src/components/sidebar/routePathView/NewRoutePathView.tsx
```diff
@@ -71,15 +73,33 @@ class NewRoutePathView extends React.Component<INewRoutePathViewProps, INewRoute
...
                 this.props.networkStore!.setSelectedTransitTypes([line.transitType]);
             }
         }
     }
 
+    public onSave = async () => {
+        this.setState({ isLoading: true });
+        try {
+            await RoutePathService.createRoutePath(this.props.routePathStore!.routePath!);
+            NotificationStore.addNotification({
```
Divinare: Our conventions say: 
Use @inject when importing stores whenever possible.

src/components/sidebar/routePathView/NewRoutePathView.tsx
```diff
@@ -88,40 +108,45 @@ class NewRoutePathView extends React.Component<INewRoutePathViewProps, INewRoute
     }
 
     public onChange = (property: string, value: string) => {
-        this.setState({
-            currentRoutePath: { ...this.state.currentRoutePath!, [property]: value },
-        });
+        this.props.routePathStore!.updateRoutePathField(property, value);
```
Divinare: updateRoutePathProperty, (we pass property named variable there) + property is used in our project rather than field

src/services/routePathService.ts
Divinare: or other option would be to have:

updateRoutePath & apliClient.updateRoutePath

createRoutePath & apiClient.createRoutePath

src/components/sidebar/routePathView/NewRoutePathView.tsx
kronken: I don't think it would be clear enough what the logic actually means, if we would use the first suggestion. We will probably need to plan how we initiate start times in the future. However, one relatively good solution would to initiate the new routepath, with a suitable starttime, would be to place the logic in the createNewRoutePathFromOld function.

Divinare: Yes, its factory is a right place for creating it if it needs some more complex logic than timestamp + 1. It can be planned in the future

Divinare:
Review ok.

Still need to test.

PR #469, opened 2018-12-12
466 Add scalebar

### LINKED ISSUE #466: Add scalebar

kronken:
Closes #466

### DIFF
--- a/src/components/map/Map.tsx
+++ b/src/components/map/Map.tsx
@@ -60,13 +60,13 @@ class LeafletMap extends React.Component<IMapProps> {
     componentDidMount() {
         const map = this.getMap();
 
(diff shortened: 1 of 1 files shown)


src/components/map/Map.tsx
```diff
@@ -60,6 +60,8 @@ class LeafletMap extends React.Component<IMapProps> {
     componentDidMount() {
         const map = this.getMap();
 
+        L.control.scale({ imperial: false }).addTo(map);
```
neodyymi: Now we use `map.addControl(Control...)` and this way mixed. Might be better to stick with one.

kronken: fixed

PR #494, opened 2019-01-02
493 Show routepath that is being edited

### LINKED ISSUE #493: Show routepath that is being edited

kronken:
Closes #493 

- UI now shows routepath that is being edited
- New tool for enabling neighbouring nodes
- Removing routepath list on map outside of routepathlist view
- Fitbounds now has padding and maxZoom which makes it a bit nicer
- Modifying information of routepath now happens in store and not state


src/components/map/layers/edit/NewRoutePathLayer.tsx
```diff
@@ -121,6 +125,17 @@ class RoutePathLayer extends Component<IRoutePathLayerProps> {
         return this.renderNode(firstNode, 0);
     }
 
+    private calculateBounds() {
```
neodyymi: arrow function for style?

src/components/map/layers/edit/NewRoutePathLayer.tsx
```diff
@@ -129,6 +144,18 @@ class RoutePathLayer extends Component<IRoutePathLayerProps> {
...
         }
+
+        if (
+            routePathStore &&
+            routePathStore!.routePath &&
+            !this.props.toolbarStore!.isSelected(ToolbarTool.AddNewRoutePath)) {
+            const bounds = this.calculateBounds();
+            if (bounds.isValid()) {
+                this.props.fitBounds(
+                    this.calculateBounds(),
```
neodyymi: Why calculate it again? Shouldn't this be same as const bounds?

src/components/map/layers/edit/NewRoutePathLayer.tsx
```diff
@@ -129,6 +144,18 @@ class RoutePathLayer extends Component<IRoutePathLayerProps> {
             && routePathStore!.neighborLinks.length === 0) {
             this.getNeighborsForExistingRoutePath();
         }
+
+        if (
+            routePathStore &&
```
neodyymi: Is this necessary?

neodyymi:
There's still some odd behaviour with zooming rather far out when creating a new routepath from old in some cases. May be an issue with small screen. That can be dealt with later.

PR #508, opened 2019-01-10
498 nodelayer nodemarker refactor

neodyymi:


src/components/map/layers/RoutePathLinkLayer.tsx
```diff
@@ -2,11 +2,11 @@ import React, { Component } from 'react';
 import * as L from 'leaflet';
 import { Polyline, FeatureGroup } from 'react-leaflet';
 import { observer, inject } from 'mobx-react';
-import { IRoutePathLink } from '~/models';
+import { INode, IRoutePathLink } from '~/models';
 import NodeType from '~/enums/nodeType';
 import { PopupStore } from '~/stores/popupStore';
-import NodeLayer from './NodeLayer';
-import NodeMarker from './NodeMarker';
+import NodeMarker from '~/components/map/layers/objects/NodeMarker';
```
kronken: These two components are from the same folder, but imported in different ways. I don't want to set rules on which way to import, but I think in these cases the fact that they are located in the same folder, could be clearer, if they were imported in the same way.

src/components/map/layers/RoutePathLinkLayer.tsx
```diff
@@ -47,29 +47,32 @@ class RoutePathLayer extends Component<RoutePathLinkLayerProps> {
     }
 
     private renderNodes() {
+        const openPopup = (node: INode) => () => {
```
kronken: Would it be possible to place this outside of the function scope? Declaring functions inside the function forces us to recreate the function, every time the parent function runs

src/components/map/layers/RoutePathLinkLayer.tsx
```diff
@@ -47,29 +47,32 @@ class RoutePathLayer extends Component<RoutePathLinkLayerProps> {
...
-        return routePathLinks.map((routePathLink, index) => {
-            return (
-                <div key={index}>
-                    <NodeLayer
-                        key={`${routePathLink.startNode.id}`}
+        const nodes = routePathLinks
+            .map((routePathLink, index) => {
+                return (
+                    <NodeMarker
+                        key={`${routePathLink.startNode.id}-${index}`}
```
kronken: What is -${index} supposed to do here? How is the startNode.id not unique enough?

src/components/map/layers/RoutePathLinkLayer.tsx
```diff
@@ -47,29 +47,32 @@ class RoutePathLayer extends Component<RoutePathLinkLayerProps> {
...
-                        />
-                    }
-                </div>
-            );
-        });
+                );
+            });
+        nodes.push(
+            <NodeMarker
+                key={`${routePathLinks[routePathLinks.length - 1]
```
kronken: Now we specify routePathLinks[routePathLinks.length - 1] three times in this block. Maybe we could create a variable named lastRoutePathLink, and use that to not duplicate fetching of the last link?

src/components/map/layers/objects/NodeMarker.tsx
```diff
@@ -0,0 +1,169 @@
+import React, { Component } from 'react';
```
kronken: The folder name 'objects' is maybe a bit too generic, or what do you think? We are using object-oriented programming, so the name 'object' can basically mean anything in my opinion.

src/components/map/layers/objects/NodeMarker.tsx
```diff
@@ -0,0 +1,169 @@
...
+        const latLng = GeometryService.iCoordinateToLatLng(this.props.node.coordinates);
+
+        const icon = createDivIcon(
+                <div className={classnames(s.nodeBase, this.getMarkerClass())}>
+                    {this.renderMarkerLabel()}
+                </div>,
+        );
+        const displayCircle = this.isSelected() && nodeType === NodeType.STOP;
+        return (
+            <>
```
kronken: Is this <> wrapper necessary?

src/components/sidebar/nodeView/NodeView.tsx
```diff
@@ -43,11 +43,11 @@ class NodeView extends React.Component
         };
     }
 
-    public componentDidMount() {
+    public async componentDidMount() {
         const selectedNodeId = this.props.match!.params.id;
         if (selectedNodeId) {
             this.props.mapStore!.setSelectedNodeId(selectedNodeId);
-            this.queryNode(selectedNodeId);
+            await this.queryNode(selectedNodeId);
```
kronken: Just out of interest, did the old code introduce problems? Or was this just modified since it is nicer? In my opinion it is a good change, so I don't question that. I just wonder if fireAndForget style executions in lifecycle methods introduce problems

src/components/map/layers/objects/NodeMarker.tsx
neodyymi: Remnant of old code.. Removing.

src/components/map/layers/RoutePathLinkLayer.tsx
neodyymi: True, I guess it's not needed

PR #521, opened 2019-01-17
495 rp calc length

### LINKED ISSUE #495: Implement functionality for calculating length of routepath

kronken:
Closes #495


src/stores/routePathStore.ts
```diff
@@ -47,7 +48,7 @@ export class RoutePathStore {
     }
 
     @action
-    updateRoutePathProperty(property: string, value: string) {
+    updateRoutePathProperty(property: string, value: any) {
```
neodyymi: value: string|number|Date ?

src/util/lengthCalculator.ts
```diff
@@ -0,0 +1,27 @@
+import * as L from 'leaflet';
+import IRoutePathLink from '../models/IRoutePathLink';
+
+const betweenCoordinates = (start: [number, number], end: [number, number]) => {
+    return L.latLng(start).distanceTo(L.latLng(end));
+};
+
+const fromPositions = (positions: [[number, number]]) => {
```
neodyymi: could also use `positions: IRoutePathLink['positions']`

src/util/lengthCalculator.ts
```diff
@@ -0,0 +1,27 @@
+import * as L from 'leaflet';
+import IRoutePathLink from '../models/IRoutePathLink';
+
+const betweenCoordinates = (start: [number, number], end: [number, number]) => {
+    return L.latLng(start).distanceTo(L.latLng(end));
```
kronken: Are the coordinates the right way?

src/util/lengthCalculator.ts
neodyymi: True.. Somehow not liking the array of arrays implementation here and maybe would rather see an array of Latlng or ICoordinates, but that would make extra work for the code asking for this information.

kronken: We could make issue of this, create some less messy logic for handling coordinates

PR #559, opened 2019-01-29
536 start end node info & styling fixed for list objects

### LINKED ISSUE #536: Make start and end node more visible under reitinlinkki

kronken:
Closes #536


src/components/sidebar/routePathView/routePathListTab/routePathListObject.scss
```diff
@@ -9,11 +9,53 @@
...
+        .linkIcon {
+            display: block;
+            height: 90px;
+            background-color: #639fd5;
+            width: 5px;
+            position: absolute;
+            top: 50%;
+            left: 50%;
+            margin-left: -2.5px;
+            margin-top: -45px;
```
neodyymi: `transform: translate(x, y)` may be a better way to do this. 
```
top 50%;
left: 50%;
margin: 0;
transform: translate(-50%, -50%)
```
should be in the middle
Now also if the height is ever changed, for example to make the link icon taller, it still stays in the middle.

src/components/sidebar/routePathView/routePathListTab/routePathListObject.scss
```diff
@@ -9,11 +9,53 @@
...
+            height: 10px;
+            width: 10px;
+            border-radius: 5px;
+            background-color: white;
+        }
 
-    .shadow {
-        background-color: #80808024;
+        .linkIcon {
+            display: block;
```
neodyymi: `display:block` is default for div

src/components/sidebar/routePathView/routePathListTab/routePathListObject.scss
kronken: thanks, fixed!

src/components/sidebar/routePathView/routePathListTab/routePathListObject.scss
kronken: fixed

PR #582, opened 2019-02-05
Enabled editing of single link + saving edited link to editNetworkStore

### LINKED ISSUE #573: Remove possibility to move nodes in linkview

### LINKED ISSUE #575: Implement link geometry editing, using geojson.io like tool

Divinare:
Closes #575 
Closes #573 

* EditNetworkLayer now supports editing of single links when user is at /link/ subSite.
   * If user is at some other site, links are drawn as before (non editable). For example if current site is /node/
   * Edited link is saved on store each time user drags a vertex

Used leaflet-editable library here: https://www.npmjs.com/package/leaflet-editable
   * API: http://leaflet.github.io/Leaflet.Editable/doc/api.html

### DIFF
--- a/src/components/map/layers/edit/EditLinkLayer.tsx
+++ b/src/components/map/layers/edit/EditLinkLayer.tsx
@@ -23,15 +23,15 @@ class EditLinkLayer extends Component<IEditLinkLayerProps> {
     private editableLinks: L.Polyline[] = [];
 
     componentDidMount() {
-        this.reactionDisposer = reaction(() =>
-        this.props.editNetworkStore!.links,
-                                         () => {
-                                             const links = this.props.editNetworkStore!.links;
-                                             if (links.length === 0) {
-                                                 this.removeOldLinks();
-                                             }
-                                         },
-            );
+        this.reactionDisposer = reaction(
+            () => this.props.editNetworkStore!.links,
+            () => {
+                const links = this.props.editNetworkStore!.links;
+                if (links.length === 0) {
+                    this.removeOldLinks();
+                }
+            },
+        );
     }
 
     private removeOldLinks = () => {
@@ -47,7 +47,7 @@ class EditLinkLayer extends Component<IEditLinkLayerProps> {
         this.reactionDisposer();
     }
 
-    private drawEditableLinks = () => {
+    private drawEditableLink = () => {
         const links = this.props.editNetworkStore!.links;
         const map = this.props.leaflet.map;
         const isLinkView = Boolean(matchPath(navigator.getPathName(), SubSites.link));
@@ -104,7 +104,7 @@ class EditLinkLayer extends Component<IEditLinkLayerProps> {
         const isLinkViewVisible = Boolean(matchPath(navigator.getPathName(), SubSites.link));
         if (!isLinkViewVisible) return null;
 
-        this.drawEditableLinks();
+        this.drawEditableLink();
 
         return (
             this.renderNodes()
--- a/src/components/map/layers/edit/EditLinkLayer.tsx
+++ b/src/components/map/layers/edit/EditLinkLayer.tsx
@@ -1,24 +1,23 @@
 import React, { Component } from 'react';
 import * as L from 'leaflet';
-import { Polyline, withLeaflet } from 'react-leaflet';
+import { withLeaflet } from 'react-leaflet';
 import { matchPath } from 'react-router';
 import { inject, observer } from 'mobx-react';
 import navigator from '~/routing/navigator';
 import SubSites from '~/routing/subSites';
 import { INode, ILink } from '~/models';
 import { EditNetworkStore } from '~/stores/editNetworkStore';
-import TransitTypeColorHelper from '~/util/transitTypeColorHelper';
 import NodeMarker from '../mapIcons/NodeMarker';
 import { LeafletContext } from '../../Map';
 
-interface IEditNetworkLayerProps {
+interface IEditLinkLayerProps {
     editNetworkStore?: EditNetworkStore;
     leaflet: LeafletContext;
 }
 
 @inject('editNetworkStore')
 @observer
-class EditNetworkLayer extends Component<IEditNetworkLayerProps> {
+class EditLinkLayer extends Component<IEditLinkLayerProps> {
     editableLinks: L.Polyline[] = [];
 
     private drawEditableLinks() {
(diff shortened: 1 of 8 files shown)


src/components/map/layers/edit/EditNetworkLayer.tsx
```diff
@@ -51,14 +98,14 @@ class EditNetworkLayer extends Component<IEditNetworkLayerProps> {
         return (
             <NodeMarker
                 key={node.id}
-                isDraggable={true}
```
kronken: NodeMarker should be moveable under node view, this changes break it.

src/components/map/layers/edit/EditNetworkLayer.tsx
```diff
@@ -1,23 +1,76 @@
...
+        map.off('editable:vertex:dragend');
+        map.on('editable:vertex:dragend', () => {
+            const currentLinks = this.props.editNetworkStore!.links.map(link => ({ ...link }));
+            const latlngs = this.editableLinks[0].getLatLngs()[0] as L.LatLng[];
+            currentLinks[0].geometry = latlngs;
+            this.props.editNetworkStore!.setLinks(currentLinks);
+        });
+    }
+
+    private drawEditableLinkToMap(link: ILink, key: number) {
```
kronken: key is not used

src/components/map/layers/edit/EditNetworkLayer.tsx
```diff
@@ -1,23 +1,76 @@
...
 
 @inject('editNetworkStore')
 @observer
 class EditNetworkLayer extends Component<IEditNetworkLayerProps> {
+    editableLinks: L.Polyline[] = [];
+
+    private drawEditableLinks() {
+        const links = this.props.editNetworkStore!.links;
+        const map = this.props.leaflet.map;
+        const isLinkView = Boolean(matchPath(navigator.getPathName(), SubSites.link));
```
kronken: We currently decide isLinkView in multiple places in the file, maybe we could set a private class variable with this data instead. Which could be set in componentDidUpdate && componentDidMount?

src/components/map/layers/edit/EditNetworkLayer.tsx
```diff
@@ -1,23 +1,76 @@
...
+                _linkLatLng.forEach((coords: any) => {
+                    const coordsToDisable = [coords[0], coords[coords.length - 1]];
+                    coordsToDisable.forEach((coordToDisable: any) => {
+                        const vertexMarker = coordToDisable.__vertex;
+                        vertexMarker.dragging.disable();
+                        vertexMarker.setOpacity(0);
+                    });
+                });
+            });
+            this.editableLinks.push(editableLink);
```
kronken: This could be simplified slightly by using:

```
   private drawEditableLinkToMap(link: ILink) {
        const map = this.props.leaflet.map;
        if (map) {
            const editableLink = L.polyline([link.geometry]).addTo(map);
            editableLink.enableEdit();
            const links = editableLink.getLatLngs() as L.LatLng[][];
            const coords = links[0];
            const coordsToDisable = [coords[0], coords[coords.length - 1]];
            coordsToDisable.forEach((coordToDisable: any) => {
                const vertexMarker = coordToDisable.__vertex;
                vertexMarker.dragging.disable();
                vertexMarker.setOpacity(0);
            });

            this.editableLinks.push(editableLink);
        }
    }
```

editable.getLatLngs() will in our case always return L.LatLngs[][]. This is a list of line sections, but in our case the line will always only be in one section. So, we can extract our line with:

```
const links = editableLink.getLatLngs() as L.LatLng[][];
const coords = links[0];
```

src/components/map/layers/edit/EditNetworkLayer.tsx
Divinare: isLinkViewVisible now being used once at render

src/components/map/layers/edit/EditNetworkLayer.tsx
Divinare: This was a nice simplification, thanks

Divinare:
Can be reviewed again.

src/components/map/layers/edit/EditLinkLayer.tsx
```diff
@@ -0,0 +1,115 @@
...
+}
+
+@inject('editNetworkStore')
+@observer
+class EditLinkLayer extends Component<IEditLinkLayerProps> {
+    private reactionDisposers: IReactionDisposer[] = [];
+    private editableLinks: L.Polyline[] = [];
+
+    componentDidMount() {
+        this.reactionDisposers.push(reaction(
```
kronken: No need for reactionDisposers to have multiple reactions.

src/components/map/layers/edit/EditLinkLayer.tsx
```diff
@@ -0,0 +1,115 @@
...
+    private removeOldLinks = () => {
+        const map = this.props.leaflet.map;
+        // Remove (possible) previously drawn links from map
+        this.editableLinks.forEach((editableLink) => {
+            map!.removeLayer(editableLink);
+        });
+        this.editableLinks = [];
+    }
+
+    componentWillUnmount() {
```
kronken: I don't find any code style rules about this currently, but I assume we usually try to put lifecycle methods at the bottom.

src/components/map/layers/edit/EditLinkLayer.tsx
```diff
@@ -0,0 +1,115 @@
...
+        this.reactionDisposers.forEach(r => r());
+    }
+
+    private drawEditableLinks = () => {
```
kronken: change method name to renderLink() or renderEditableLink() ?

src/components/map/layers/edit/EditNodeLayer.tsx
```diff
@@ -1,23 +1,29 @@
...
-import NodeLocationType from '~/types/NodeLocationType';
 import { EditNetworkStore } from '~/stores/editNetworkStore';
 import TransitTypeColorHelper from '~/util/transitTypeColorHelper';
 import NodeMarker from '../mapIcons/NodeMarker';
+import { LeafletContext } from '../../Map';
 
-interface IEditNetworkLayerProps {
+interface IEditNodeLayerProps {
     editNetworkStore?: EditNetworkStore;
+    leaflet: LeafletContext;
```
kronken: leaflet props is not used

src/components/map/layers/edit/EditNodeLayer.tsx
```diff
@@ -68,4 +70,4 @@ class EditNetworkLayer extends Component<IEditNetworkLayerProps> {
     }
 }
 
-export default EditNetworkLayer;
+export default withLeaflet(EditNodeLayer);
```
kronken: withLeaflet is not needed, since we dont use leaflet in this class

src/components/map/layers/edit/EditNodeLayer.tsx
```diff
@@ -34,12 +40,6 @@ class EditNetworkLayer extends Component<IEditNetworkLayerProps> {
         );
     }
 
-    private onMoveMarker = (node: INode) =>
```
kronken: onMoveMarker is required to update the store with new coordinate values of the node

src/components/map/layers/edit/EditLinkLayer.tsx
kronken: This seems to be this way in other places as well, so no need to update this if you don't want to.

Divinare:
Fixes done

src/components/map/layers/edit/EditLinkLayer.tsx
Divinare: Using one reactionDisposer now

src/components/map/layers/edit/EditNodeLayer.tsx
Divinare: Fixed

src/components/map/layers/edit/EditNodeLayer.tsx
Divinare: Good point

PR #600, opened 2019-02-08
Fix rp save

### LINKED ISSUE #585: RoutePathView is never warning of unsaved modifications

### LINKED ISSUE #591: RoutePath save functionality
NOTE: make these changes to fix-rp-save branch

* Previously save button validation worked at 'Reitinsuunnan tiedot' tab. Code moved to parent component (look at https://github.com/HSLdevcom/jore-map-ui/pull/590) so that the same save button would work for also 'Solmut ja linkit' tab.

Divinare:
Closes #591 
Closes #585 

* Fixed isDirty check, saveButton should be green/grey depending on if modifications has been made or not
   * done with lodash: _.isEqual()
   * Also there was a bug in markInvalidFields() method that I fixed
* I had a problem: inputFields needed to be refreshed when user presses "Muokkaa/Peruuta" button
   * solution: used force refresh (if inputContainer's disabled prop changed, validation was forced)

How to test:
Try to edit old and new routePaths, try to see if everything related to validating works


src/components/sidebar/routePathView/RoutePathView.tsx
```diff
@@ -107,23 +156,41 @@ class RoutePathView extends React.Component<IRoutePathViewProps, IRoutePathViewS
...
     }
 
-    async componentDidMount() {
-        if (this.props.isAddingNew) {
-            await this.initializeAsAddingNew();
-        } else {
-            await this.fetchRoutePath();
+    private isFormValid = () => {
+        return !Object.values(this.state.invalidFieldsMap)
+            .some(fieldIsValid => !fieldIsValid);
```
kronken: Maybe change this to:
return Object.values(this.state.invalidFieldsMap).every(fieldIsValid => fieldIsValid).

src/components/sidebar/routePathView/RoutePathView.tsx
```diff
@@ -135,16 +202,30 @@ class RoutePathView extends React.Component<IRoutePathViewProps, IRoutePathViewS
             );
         }
         if (!this.props.routePathStore!.routePath) return null;
+
+        const isSaveButtonDisabled = this.state.isEditingDisabled
+        || !this.props.routePathStore!.isDirty
```
kronken: indent?

src/components/sidebar/routePathView/RoutePathView.tsx
```diff
@@ -135,16 +202,30 @@ class RoutePathView extends React.Component<IRoutePathViewProps, IRoutePathViewS
...
+        || !this.props.routePathStore!.isGeometryValid
+        || !this.isFormValid();
+
         return (
             <div className={s.routePathView}>
                 <RoutePathHeader
+                    hasModifications={this.props.routePathStore!.isDirty}
                     routePath={this.props.routePathStore!.routePath!}
-                    isAddingNew={this.props.isAddingNew}
+                    isNewRoutePath={this.props.isNewRoutePath}
```
kronken: Sometime we use this.props.isNewRoutePath and sometimes this.routePathIsNew(), chose one way and use that consistently. Personally I think

```
   private routePathIsNew = () => {
        return this.props.isNewRoutePath;
    }
```

function could be removed and just use the prop directly. 

this.props.isNewRoutePath is not exactly much longer than this.routePathIsNew()

src/components/sidebar/routePathView/RoutePathView.tsx
kronken: Hmm, this is actually faster this way.

PR #606, opened 2019-02-08
Dialog store & openFadeMessage implemented

### LINKED ISSUE #593: Notification fade dialog
An ideal solution would be to say:

import alert.ts

alert.openFadeDialog('Tallennettu')

Divinare:
Closes #593

How to test:

Save routePath, you should see fadeDialog appear (time visible in development mode: 500ms)

### DIFF
--- a/src/stores/dialogStore.ts
+++ b/src/stores/dialogStore.ts
@@ -21,9 +21,10 @@ export class DialogStore {
     @action
     public setFadeMessage = (message: string) => {
         this._message = message;
-        setTimeout(() => {
-            this.close();
-        },         Constants.FADE_DIALOG_TIMEOUT);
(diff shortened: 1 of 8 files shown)


src/stores/dialogStore.ts
```diff
@@ -0,0 +1,37 @@
...
+    get isDialogOpen(): boolean {
+        return this._message !== null;
+    }
+
+    @action
+    public setFadeMessage = (message: string) => {
+        this._message = message;
+        setTimeout(() => {
+            this.close();
+        },         Constants.FADE_DIALOG_TIMEOUT);
```
kronken: we could probably fix this indent by moving the start of the arrow function () to a new row.

PR #610, opened 2019-02-11
595 Refactoring error handling

### LINKED ISSUE #595: Showing error messages on map with ExceptionCatcher
Create ExceptionCatcher.ts that can catch errors from components. For example, in RoutePathView.tsx we have:

throw new Error(`Route not found, routeId: ${queryParams.routeId}`);

This should be caught by ExceptionCatcher and error should be visible with red on map

kronken:
Closes #595 

Easiest way to test is to open and close login view.
Or for more realistic test: update api url to be broken, and see error messages when UI cant reach BE.


src/components/shared/searchView/SearchResults.tsx
```diff
@@ -71,16 +73,20 @@ class SearchResults extends React.Component<ISearchResultsProps, ISearchResultsS
     }
 
     private fetchAllLines = async () => {
-        const lines = await LineService.fetchAllLines();
-        if (lines !== null) {
+        try {
+            const lines = await LineService.fetchAllLines();
```
Divinare: Hmm, now if lines is empty array, it will not throw error.

Maybe we need something like:
const lines = await LineService.fetchAllLines();
if (lines.length === 0) throw new Error();

src/components/sidebar/routePathView/RoutePathView.tsx
```diff
@@ -73,13 +72,20 @@ class RoutePathView extends React.Component<IRoutePathViewProps, IRoutePathViewS
...
+                this.props.routePathStore!.setRoutePath(await this.createNewRoutePath());
+            } else {
+                this.props.routePathStore!.setRoutePath(
+                    RoutePathFactory.createNewRoutePathFromOld(
+                        this.props.routePathStore!.routePath!,
+                    ),
+                );
+            }
+            this.props.toolbarStore!.selectTool(ToolbarTool.AddNewRoutePathLink);
+        } catch (ex) {
```
Divinare: I have used catch (e) before. What ex refers to?

src/components/sidebar/routePathView/RoutePathView.tsx
```diff
@@ -96,28 +102,32 @@ class RoutePathView extends React.Component<IRoutePathViewProps, IRoutePathViewS
...
     private setTransitType = async () => {
         const routePath = this.props.routePathStore!.routePath;
         if (routePath && routePath.lineId) {
-            const line = await LineService.fetchLine(routePath.lineId);
-            if (line) {
+            try {
+                const line = await LineService.fetchLine(routePath.lineId);
                 this.props.networkStore!.setSelectedTransitTypes([line.transitType]);
+            } catch (ex) {
+                this.props.errorStore!.push('Linja haku ei onnistunut');
```
Divinare: Linjan haku ei...

src/services/geometryService.ts
```diff
@@ -1,7 +1,7 @@
 import proj4 from 'proj4';
 import CoordinateSystem from '~/enums/coordinateSystem';
 
-class GeometryService {
+class GeometryService { // TODO: Move out of /services
```
Divinare: Good comment. I have also noticed that this should be moved...

src/components/shared/searchView/SearchResults.tsx
kronken: Why would we throw an error if fetchAllLines return empty array?
fetchAllLines should throw an error if it has problems fetching something, it should never return empty array if an error occurs. Only time it should return empty array is if the database is empty, and in that case, it should not throw any error

src/components/sidebar/routePathView/RoutePathView.tsx
kronken: Both 'e' and 'ex' refers to exception, and both are widely used globally. But could change it to 'e' to keep it consistent within our codebase.

Divinare:
Testing done.

Try to input something when entering:
http://localhost:3000/routes/?routes[0]=1234

This causes error:
this.paginatedDiv.current!.scrollTo(0, 0);

Sidebar looks broken:
http://localhost:3000/routePath/0012,2011-01-01T00:00:00,1

White box appears even before the error is shown:
http://localhost:3000/node/1170009/

Hmm...
Try this with small & wide screen:
http://localhost:3000/link/2301214,1302216,1/?routes[0]=1234J

Solution something like:

white-space: nowrap; 
overflow: hidden;
text-overflow: ellipsis; (this should put ... to the end, not working for some reason)

PR #622, opened 2019-02-14
574 edit link

### LINKED ISSUE #574: Save changes to LinkView

kronken:
Closes #574


src/components/controls/Checkbox.tsx
```diff
@@ -4,31 +4,33 @@ import * as s from './checkbox.scss';
...
+                onClick={this.onChange}
                 className={s.container}
             >
                 {this.props.text}
                 <input
-                    type={'checkbox'}
+                    type='checkbox'
                     checked={this.props.checked}
-                    onChange={this.doNothing}
+                    onChange={this.onClick}
```
Divinare: Why were these changes made? I don't see any checkboxes in the scope of this PR (linkView)

src/components/controls/Dropdown.tsx
```diff
@@ -10,7 +10,7 @@ interface IDropdownBaseProps {
     label?: string;
     selected: string;
     disabled?: boolean;
-    onChange(selectedItem: string): void;
+    onChange?: (selectedItem: string) => void;
```
Divinare: Why onChange is optional? Shouldn't dropdown always require onChange?

src/components/controls/TransitToggleButton.tsx
```diff
@@ -30,15 +31,16 @@ class TransitToggleButton extends React.Component
 
     public render() {
         return (
-            <button
+            <div
                 className={classNames(
                     s.button,
                     this.getToggledButtonClass(this.props.type, this.props.toggled),
+                    !this.props.disabled ? s.active : undefined,
```
Divinare: A bad practice is to have active as a "special case". Here active means "normal state"

Instead, add cursor: pointer to .button and use code:

```
this.props.disabled ? s.disabled : undefined

.disabled {
     pointer-events: none;
     cursor: none;
}
```

src/components/controls/TransitToggleButton.tsx
```diff
@@ -18,7 +19,7 @@ interface ITransitToggleButtonState {
 class TransitToggleButton extends React.Component
   <ITransitToggleButtonProps, ITransitToggleButtonState> {
     public toggleActivity = () => {
-        this.props.toggleActivity(this.props.type);
+        !this.props.disabled && this.props.toggleActivity(this.props.type);
```
Divinare: This shouldn't be needed if you look at my other comment

src/components/controls/TransitToggleButtonBar.tsx
```diff
@@ -8,6 +8,7 @@ interface ITtransitToggleButtonBarProps {
     selectedTransitTypes: TransitType[];
     toggleSelectedTransitType?: (type: TransitType) => void;
     disabled?: boolean;
+    blurred?: boolean;
```
Divinare: I'm not convinced you need both disabled & blurred

If you use pointer-events: none; (at TransitToggleButton.tsx), these changes can be reverted I think

src/components/controls/TransitToggleButtonBar.tsx
```diff
@@ -25,28 +26,33 @@ class TransitToggleButtonBar extends React.Component<ITtransitToggleButtonBarPro
...
+                    disabled={this.props.disabled}
                 />
                 <TransitToggleButton
                     toggleActivity={this.toggleType}
                     toggled={this.props.selectedTransitTypes.includes(TransitType.FERRY)}
                     type={TransitType.FERRY}
+                    disabled={this.props.disabled}
                 />
-                { this.props.disabled &&
+                { this.props.blurred &&
```
Divinare: If blurred puts on s.disabledOverlay --> rename as s.blurredOverlay

src/components/controls/checkbox.scss
```diff
@@ -23,7 +24,7 @@
 /* Create a custom checkbox */
 .checkmark {
     position: absolute;
-    top: 0;
+    top: 10px;
```
Divinare: MapLayers checkboxes are placed slightly off, is top: 10px needed?

src/components/controls/transitToggleButton.scss
```diff
@@ -1,11 +1,16 @@
...
     border: hidden;
-    cursor: pointer;
     transition: .2s;
+    display: flex;
+    align-items: center;
+    justify-content: center;
+}
+
+.active {
+    cursor: pointer;
```
Divinare: I would put cursor: pointer back to .button as its the default state

src/components/sidebar/routePathView/routePathListTab/RoutePathListLink.tsx
```diff
@@ -52,6 +55,79 @@ class RoutePathListLink extends React.Component<IRoutePathListLinkProps> {
...
+                            text={'Ohitusaika kirja-aikat.'}
+                            onClick={this.onChange}
+                        />
+                    </div>
+
+                </div>
+                <div className={s.flexRow}>
+                    <Dropdown
+                        label='SOLMU HASTUS-PAIKKANA'
+                        items={['Kyllä', 'Ei']}
```
Divinare: Hmm, need to check this but should this be a key: value array? Maybe a "yesNoCodeList" could be used here.

src/components/controls/Dropdown.tsx
kronken: We have permanent disabled dropbox, that we necessary don't want to specify onChange methods for.

src/components/controls/TransitToggleButtonBar.tsx
kronken: They are different things. In the start view we want the transittoggles to be both disabled, but also blurred so people understand that it is not active. This is different from what we want in link view, in link view we want to show the transit type for the user, so we want disabled toggles, but not blurred.

src/components/sidebar/routePathView/routePathListTab/RoutePathListLink.tsx
kronken: The problem with yesNoCodeList, is that it reflects a boolean, and you can not have booleans as keys in a codelist. We could add a new component for boolean values, that we could use in other places as well.

kronken: https://github.com/HSLdevcom/jore-map-ui/issues/624

src/components/controls/Checkbox.tsx
kronken: Checkbox was used in linkview, but it was now moved to reitinsuunta list. I noticed the component had many flaws, so I improved it a bit. onClickHandlers was created in the render function, onClick required event parameter, render function had specified it returns any, curly braces around string in JSX.

kronken: I however reverted the onChange function, since that seems to not be working as I planned it would

src/components/controls/TransitToggleButtonBar.tsx
kronken: Why would the parents need to know it is an overlay?

src/components/controls/checkbox.scss
kronken: The checkboxes are broken in master, they practically only work if the text in only on one row. This was a bad attempt to fix it. I reverted the change, so they are now back to normal, but still broken.

src/components/controls/TransitToggleButtonBar.tsx
kronken: missunderstood your comment, this is now fixed.

PR #628, opened 2019-02-18
625 link direction visualization

### LINKED ISSUE #625: Link direction visualization
bbecquet/Leaflet.PolylineDecorator

kronken:
Added direction arrows on RoutesView, RoutePathView, LinkView and NodeView.

Closes #625


Divinare:
Testing:

Bug 1:

Try to open page: http://localhost:3000/routePath/0033,2019-02-06T08:44:08,1/?routes[0]=0033

--> an error is shown in console

The same happens if opening:

http://localhost:3000/node/4170009/?routes[0]=0035
or
http://localhost:3000/node/4170245/?routes[0]=0035


Hmm, now arrows are always on top of routePathLinks. If two routePaths are on top of each other, arrows are shown on top all the time. But maybe that is fine.

Bug 2:

Highlighting routePath doesn't always put it on top of all other visible routePaths

src/components/map/layers/RoutePathLinkLayer.tsx
```diff
@@ -85,14 +95,33 @@ class RoutePathLinkLayer extends Component<RoutePathLinkLayerProps> {
...
+
+        return createCoherentLinesFromPolylines(geoms)
+            .map((geom, index) => (
+                <ArrowDecorator
+                    key={index}
+                    color={this.props.color}
+                    geometry={geom}
+                    onClick={this.props.onClick(this.layerRef)}
+                />
+            ),
```
Divinare: What createCoherentLinesFromPolylines does and why? I guess there must be an easier way without that helper function

Much clearner way would be to say:
```
return this.props.routePathLinks.map(routePathLink) =>
    routePathLink.geometry.map(polyline, index) => (
                <ArrowDecorator
                    key={index}
                    color={this.props.color}
                    geometry={geom}
                    onClick={this.props.onClick(this.layerRef)}
                />
           )
```
Would code like this work?

src/components/map/layers/ArrowDecorator.ts
```diff
@@ -0,0 +1,69 @@
...
+import L, { PolylineDecorator, LatLng } from 'leaflet';
+import { LeafletContext } from '../Map';
+
+interface IArrowDecoratorProps extends PathProps {
+    leaflet: LeafletContext;
+    geometry: LatLng[];
+    onClick?: () => void;
+    color: string;
+    disableOnEventName?: string;
+    enableOnEventName?: string;
```
Divinare: Hmm, now when I think maybe "showOnEventName and hideOnEventName" would be better variable names

src/components/map/layers/RoutePathLinkLayer.tsx
kronken: It would work, but this creates arrow decoration from a set of polylines, not from one complete polyline. Having decorations made from a set of polylines makes the distance between the arrows inconsistent. Having one single polyline makes the arrows be with consistent distance between each other, throughout the whole coherent polyline.

Divinare: Add comment above the method?
/**
    Creates coherent polylines from multiple polylines: if there is a hole inside a polyline, it splits the polyline into two polylines etc.
*/
const createCoherentLinesFromPolylines = ...

kronken: I need to change the comment a bit, the polylines don't have holes. There might be holes between the polylines, and this is where we should split the final polyline.

PR #635, opened 2019-02-18
Undo and redo now works in routePath and linkView. Removed GeometryEventStore, Removed EventLog

### LINKED ISSUE #577: Fix event logging -> re implementation undo / redo of routePaths

Divinare:
Closes #577

* RoutePath undo fixed, redo implemented
* LinkView undo / redo implemented

* In routePathView, neighbor fetch is broken in undo/redo --> #636
* Implement common undo/redo store & undo/redo to nodeView:  #637
* Removed old geometryEventStore (we don't need it anymore)
* Removed eventLog (we don't need it anymore)

How to test:
* Try to add / remove routePathLinks and press undo / redo in every combination
    * same when editing link at linkView


src/components/map/toolbar/ToolbarCommonButtons.tsx
```diff
@@ -46,11 +50,19 @@ class ToolbarCommonButtons extends React.Component {
...
+                        isDisabled={false}
+                        label='Kumoa'
                     >
                         <IoMdUndo />
                     </MapControlButton>
+                    <MapControlButton
+                        onClick={this.redo}
+                        isActive={false}
+                        isDisabled={false}
+                        label='Tee uudestaan'
```
kronken: 'Tee uudestaan'? Is that actually the correct term in finnish?

src/stores/linkStore.ts
```diff
@@ -27,9 +35,34 @@ export class LinkStore {
...
     }
 
+    @action
+    public updateLinkGeometry = (latLngs: L.LatLng[]) => {
+        if (!this._link) return;
+
+        const updatedLink = _.cloneDeep(this._link);
+        updatedLink.geometry = latLngs;
+
+        const currentUndoObject = {
```
kronken: Would suggest typing to check that object complies with what we are creating.
const currentUndoObject: UndoObject = {
}

src/stores/routePathStore.ts
```diff
@@ -20,16 +19,25 @@ export enum RoutePathViewTab {
...
+}
+
 export class RoutePathStore {
     @observable private _routePath: IRoutePath|null;
     @observable private _oldRoutePath: IRoutePath|null;
     @observable private _isGeometryValid: boolean;
     @observable private _neighborRoutePathLinks: IRoutePathLink[];
+    // TODO: Move out of store to state, rename as highlightedMapObject
     @observable private _highlightedObject: string | null;
+    // TODO: Move out of store to state, rename as extendedListItems:
```
kronken: Regarding comments, I agree that we could change this. but be consistent in the naming highlightedObjects and extendedObjects points to the same set of items, so we should use consistent naming. Not highlightedMapObjects and extendedListItems => items and objects.

src/util/EventManager.ts
```diff
@@ -0,0 +1,25 @@
+
+class EventManager {
+
+    public trigger(eventName: string) {
+        const event = new Event(eventName);
+        document.dispatchEvent(event);
+    }
+
+    // TODO: use eventName: type
```
kronken: Why not add typings already now, that is practically one line?

src/components/map/toolbar/ToolbarCommonButtons.tsx
Divinare: Didn't figure out anything better. Found it from here: https://www.sanakirja.org/search.php?q=redo&l=3&l2=17

src/stores/linkStore.ts
Divinare: :+1:

src/stores/routePathStore.ts
Divinare: :+1: , changed highlightedMapObjects to highlightedMapItems. It was a bad name anyway because the type was string, not object

src/util/EventManager.ts
Divinare: I had to do this in this task but somehow forgot, good that you noticed

Divinare:
Review fixes done

That text selection was a problem for every other mapControlButton as well but it is fixed now.

PR #641, opened 2019-02-19
630 new login ui

### LINKED ISSUE #630: Login UI
* when the app starts, it shows login page /login/
   * later will direct straigtly to the app if token is valid
* clicking login button redirects user to home view
* when in dev mode, show app directly, in production show login page first(?)

Divinare:
Closes #630 

* Actual login will be done later


src/components/PrivateRoute.tsx
```diff
@@ -0,0 +1,32 @@
+import * as React from 'react';
+import { Redirect, Route, RouteComponentProps, RouteProps } from 'react-router-dom';
+import LoginStore from '~/stores/loginStore';
+
+type RouteComponent = React.FunctionComponent<RouteComponentProps<{}>> | React.ComponentClass<any>;
+
+/* tslint:disable:variable-name */
```
kronken: Why disable this? I suppose since variables are required to be in camel-case, but a better solution would maybe be to edit the tslint config, or to comply with its rules

src/components/PrivateRoute.tsx
```diff
@@ -0,0 +1,32 @@
+import * as React from 'react';
```
kronken: Any reason to use this instead of just: import React from 'react'?

src/components/login/Login.tsx
```diff
@@ -1,70 +1,69 @@
...
-            <div className={s.modalButtonBar}>
-                <Button
-                    onClick={this.closeLoginModal}
-                    type={ButtonType.SQUARE}
+                    <label className={s.checkboxContainer}>
+                        <input className={s.checkbox} type='checkbox'/>
+                        Muista minut
+                    </label>
+                </form>
+                <div
```
kronken: The login button should probably be a form submit, and be part of the form

src/components/login/Login.tsx
```diff
@@ -1,70 +1,69 @@
 import React from 'react';
 import { inject, observer } from 'mobx-react';
+import { Redirect } from 'react-router-dom';
+import { Location } from 'history';
+import hslLogo from '~/assets/hsl-logo.png';
+import loginBackground from '~/assets/loginBackground.jpg';
```
kronken: The login picture is 2500 px wide, we could probably make it a bit more optimized by having a slightly smaller image. The image is so blurry, so I don't think anyone notices if it would have one fourth of the current resolution.

src/components/login/Login.tsx
```diff
@@ -1,70 +1,69 @@
...
+        }
+
         return (
         <div className={s.loginView}>
-            <h2>Kirjaudu sisään</h2>
-            <form>
-                <label className={s.label}>
-                    Tunnus
-                <br/>
+            <img className={s.backgroundImage} src={loginBackground} />
```
kronken: Would suggest using background image instead of adding the image like this. Having background images will allow us to use properties such as 'background-size: contain', which will prevent squeezing the image on screen sizes.

Divinare:
Hmm, login window should be hidden by default in dev mode I think. So that we don't have to click "Kirjaudu" each time we refresh the page.

src/components/login/login.scss
```diff
@@ -1,25 +1,116 @@
...
+            position: relative;
+            margin-top: 20px;
+            cursor: pointer;
+            font-weight: 400;
+            white-space: nowrap;
+            font-size: $mediumFontSize;
+            border: hidden;
+            border-radius: 2px;
+            background-color: #ffffffe6;
+            color: $fontColorDefault;
```
kronken: Quite many of these properties are not needed: at least on my machine both: position, font-weight, white-space and border has these properties by default. except for border, but not needed to specify the border as hidden, when we have not added any border for it.

src/services/authService.ts
```diff
@@ -0,0 +1,12 @@
+class AuthService {
+    public static authenticate(cb: any) {
```
kronken: maybe rename cb as callback, would be slightly clearer, and we know the type of it, at least that it is a function.

src/components/PrivateRoute.tsx
Divinare: Added tslint rule

src/components/PrivateRoute.tsx
Divinare: Just copied it from somewhere... changed now

src/components/login/Login.tsx
Divinare: Resized now to 616 x 322, size dropped about 50%

src/components/login/login.scss
Divinare: Some css taken out

src/components/login/Login.tsx
Divinare: Should be fixed now

Divinare:
Review fixes done.

How to test:
change development constants --> isLoginRequired: true

PR #649, opened 2019-02-21
Generic undoStore && undo to nodeView

Divinare:
Many things happened here:
* common undoStore
    * being used at each routePathStore & link/node stores

* editNodeLayer:
   * had to make modifications to fetch nodes & links at once to init store with them in one action
   * save now resets undo status
   * editing links resets undo

* Created a new card #654 to prevent editing geometries if form is not in editing state

Test that undo still works in existing views. Test especially that undo works in nodeView

Maybe review one commit at a time to see which changes relates to which


src/stores/undoStore.ts
```diff
@@ -0,0 +1,42 @@
...
+
+    constructor() {
+        this.clear();
+    }
+
+    public addUndoObject = (undoObject: UndoObject) => {
+        // Remove the history of undo's because current state is changed
+        this._undoObjects.splice(this._undoIndex + 1);
+
+        // Insert current undoObject to the pile
```
kronken: Would suggest the word stack, a stack is ordered, a pile is not necessarily ordered.

src/stores/linkStore.ts
```diff
@@ -35,10 +34,11 @@ export class LinkStore {
     @action
     public setLink = (link: ILink) => {
         this._link = link;
-        this._undoObjects = [{
+        const undoObject: UndoObject = {
```
kronken: Would it not be more natural to use the words linkGeometryState: LinkGeometryState. UndoObject feels a bit abstract.

Then we could have syntax such as:
GeometryUndoStack.addItemToUndoStack(linkGeometryState);
instead of 
UndoStore.addUndoObject(undoObject).

src/stores/undoStore.ts
```diff
@@ -0,0 +1,42 @@
+
```
kronken: what about undoStack?
Feels weird using the word store, since we have been exclusively been using the word store as a mobx stores.

src/stores/nodeStore.ts
```diff
@@ -27,37 +38,69 @@ export class NodeStore {
...
+        this._links = newLinks;
+        this._oldLinks = newLinks;
     }
 
     @action
-    public setNode = (node: INode) => {
-        this._node = node;
-        this.setOldNode(node);
+    public changeLinkGeometry = (latLngs: L.LatLng[], index: number) => {
+        if (!this._node) throw new Error('Node was null'); // Sanity check
```
kronken: is this comment crucial?

src/stores/linkStore.ts
Divinare: Lets go with: this._undoStore.addItem(currentState);

src/stores/nodeStore.ts
Divinare: Using "should never occur"

Divinare:
Review fixes done ^

PR #651, opened 2019-02-21
634 only one location for non stop node

### LINKED ISSUE #634: Nodes that are not stops do only have one location.

kronken:
Closes #634 

Test by opening node view and edit something that is not a stop


src/stores/nodeStore.ts
```diff
@@ -64,7 +57,25 @@ export class NodeStore {
...
+            ...this._node!,
+            [property]: value,
+        };
+
+        if (this._node.type !== NodeType.STOP) {
+            if (property === 'coordinates' || property === 'type') {
+                this._node.coordinatesProjection = this._node.coordinates;
+                this._node.coordinatesManual = this._node.coordinates;
+            }
+        }
```
Divinare: Hmm, how about having updateNodeCoordinates method that has this if logic?

Or have EditNodeLayer call updateNode 3 times with:
if(NodeType !== NodeType.STOP) {
   this.props.nodeStore!.updateNode('coordinates', coordinates);
   this.props.nodeStore!.updateNode('coordinatesManual', coordinates);
   this.props.nodeStore!.updateNode('coordinatesProjection', coordinates);
}

Just trying to go without having 2 x specific if clauses in a generic function

kronken: I would like this logic to stay in the store, I don't think it is the EditNodeLayer:s responsibility to know how to update locations. We could extract the code to a separate function, but the most reasonable function I can come up with would look like this:

```
    @action
    private copyLocationForNonStops = (coordinate: LatLng) => {
        this._node!.coordinatesManual = coordinate;
        this._node!.coordinatesProjection = coordinate;
    }
```

Divinare: Actually I'm going to create a method "updateNodeGeometry" when implementing undo at #637.

Wait for that task to be merged and then move that if to that updateNodeGeometry method?

Divinare: Reason for separate updateNodeGeometry() method is that I will have to put geometries of links and node to undoStore

Divinare: Lets move that if in #637 and merge this now.

PR #679, opened 2019-03-05
646 login ui part 2

### LINKED ISSUE #646: Show user information, and logout functionality in UI

kronken:
Closes #646

- Add navigation bar
- Prevent editing fields and geometry when user doesn't have write access
- Enabling clearing of login store using logout button

Full logout will be implemented in:
https://github.com/HSLdevcom/jore-map-ui/issues/669

Save button hiding will be implemented in:
https://github.com/HSLdevcom/jore-map-ui/issues/676

### DIFF
--- a/src/stores/loginStore.ts
+++ b/src/stores/loginStore.ts
@@ -9,17 +9,17 @@ export class LoginStore {
     @observable private _userEmail?: string;
 
     constructor() {
-        this._isAuthenticated = !Constants.isLoginRequired;
+        this.clear();
     }
 
     @computed
     get isAuthenticated() {
-        return this._isAuthenticated;
+        return this._isAuthenticated || !Constants.isLoginRequired;
     }
 
     @computed
     get hasWriteAccess() {
-        return this._hasWriteAccess;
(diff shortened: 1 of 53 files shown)


src/stores/loginStore.ts
```diff
@@ -32,10 +32,13 @@ export class LoginStore {
...
+    }
 
-        // TODO: this is only temporary:
-        // tslint:disable-next-line:no-console
-        console.log(`User: ${authRespose.email} is now authenticated`);
+    @action
+    public clear() {
+        this._isAuthenticated = false;
+        this._userEmail = undefined;
+        this._hasWriteAccess = false;
```
Divinare: Hmm I wouldn't make clear func since it is only called once. I would go with:

constructor() {
      this._isAuthenticated = !Constants.isLoginRequired;
      this._hasWriteAccess = !Constants.isLoginRequired;
      this._userEmail = undefined;
}

kronken: Lets go with current implementation until we do full logout

PR #683, opened 2019-03-07
636 route path view neighbor fetch in undo

### LINKED ISSUE #636: Fix routePathView's neighbor fetch in undo/redo
* Highlight clickable nodes after undo / redo
* Fix bugs related to extend routePath tool

Divinare:
Closes #636 

* Renamed AddNewRoutePathLinkTool -> ExtendRoutePathTool
* Undo with ExtendRoutePathTool now highlights clickable nodes automatically
* Fixed bug when ArrowDecarator didn't show arrows on a link in some cases
* Fixed onClick & highlight bugs when modifiying routePath
* Refactored neighbor fetch related things in routePathStore & upsertRoutePathLayer


src/components/map/tools/ExtendRoutePathTool.ts
```diff
@@ -0,0 +1,62 @@
...
+
+    private reset() {
+        RoutePathStore.setNeighborRoutePathLinks([]);
+    }
+
+    public onNetworkNodeClick = async (clickEvent: any) => {
+        try {
+            if (!this.isNetworkNodesInteractive()) return;
+
+            const properties =  clickEvent.sourceTarget.properties;
```
kronken: nip: there are two spaces between = and clickEvent.

src/stores/routePathStore.ts
```diff
@@ -1,18 +1,14 @@
...
-    SetTargetLocation,
-    AddLinks,
-}
-
-export enum AddLinkDirection {
-    BeforeNode,
-    AfterNode,
+// Is the neighbor to add either startNode or endNode
+export enum NeighborToAddType {
+    StartNode,
```
kronken: In my opinion the old beforeNode and afterNode is more declarative. But we could figure out a third option also, since I don't really understand BeforeNode and AfterNode at the moment

src/components/map/tools/ExtendRoutePathTool.ts
Divinare: Nit* (nitpicking)

src/stores/routePathStore.ts
Divinare: I think as a nodeType (NeighborToAddType) StartNode / EndNode fits better because in our ILink we have startNode & endNode so I would go with this naming

Issue #653, opened 2019-02-22
Redirect user back to origin after login


kronken:

Divinare:
To implement this:

1. user goes http://localhost:3000/routes/?routes[0]=0033
2. save that link (http://localhost:3000/routes/?routes[0]=0033)
3. user is redirected to /login
4. put that saved link to this url:
window.location.replace(
`https://hslid-uat.cinfra.fi/openid/auth?client_id=6549375356227079&redirect_uri=${constants.AFTER_LOGIN_URL}&response_type=code&scope=email+https://oneportal.trivore.com/scope/groups.readonly`
)
5. when user gets redirected back to our app, get the saved link from afterLogin url and redirect to that url

look at my comment here: https://github.com/HSLdevcom/jore-map-ui/pull/687/files#diff-c8a101915dc6871f1f3d7ad2208d8655R67

add that saved link to that redirect call

PR #687, opened 2019-03-07
667 remember session

### LINKED ISSUE #667: Authentication: remember session if page is refreshed
* At backend we have server.ts cookie maxAge configured, user shouldn't have to log in if that cookie is alive
* Frontend should should check if session exists, if not, redirect to hslid login

### LINKED ISSUE #669: Remove session via backend on logout

Divinare:
Closes #667
Closes #669 

Closes Backend PR: https://gitlab.hsl.fi/jore/jore-map-backend/merge_requests/30

* Refresh page should work, it can even keep the same site that was refreshed
* Logout should work

* Removed PrivateRoute because it is no longer needed

* Still if user has no existing session and he goes localhost:3000/routes/.... -> /login is redirected and then /home is redirected: should be done here #653


src/components/App.tsx
```diff
@@ -29,6 +30,20 @@ interface IAppProps extends RouteComponentProps<any> {
 @inject('mapStore', 'loginStore')
 @observer
 class App extends React.Component<IAppProps, IAppState> {
+    async componentWillMount() {
+        const pathName = navigator.getPathName();
```
kronken: Extract this to new function, currently it is difficult to fast know what it does, when the function name is just componentWillMount()

src/components/App.tsx
```diff
@@ -29,6 +30,20 @@ interface IAppProps extends RouteComponentProps<any> {
...
+        if (!pathName.includes(SubSites.afterLogin)) {
```
kronken: Use standard react-router matchPath() method instead of doing it manually. Current implementation only checks if the url contains the 'afterLogin' anywhere in the url. matchPath checks this more properly and less risk of missmatches.

src/components/App.tsx
```diff
@@ -29,6 +30,20 @@ interface IAppProps extends RouteComponentProps<any> {
...
+            const response = (await ApiClient
+                .postRequest(endpoints.EXISTING_SESSION, {}) as IAuthorizationResponse);
```
kronken: Wouldn't it be better to set postRequest to take a optional parameter body? Then we could use it like:
postRequest(endpoints.EXISTING_SESSION)
postRequest(endPoints.LINK, link)

src/components/App.tsx
```diff
@@ -29,6 +30,20 @@ interface IAppProps extends RouteComponentProps<any> {
...
+            const { history } = this.props;
+            if (response.isOk) {
+                this.props.loginStore!.setAuthenticationInfo(response);
+            } else {
+                history.push(SubSites.login);
```
kronken: Use navigator.goTo(subSites.login)?

src/components/App.tsx
```diff
@@ -29,6 +30,20 @@ interface IAppProps extends RouteComponentProps<any> {
...
+            }
+        }
+        this.props.loginStore!.setIsLoginInProgress(false);
```
kronken: Why is this in store and not state? 
This field is only used in this component, component isolated observables should be in state.

src/stores/loginStore.ts
```diff
@@ -7,19 +7,26 @@ export class LoginStore {
     @observable private _isAuthenticated: boolean;
     @observable private _hasWriteAccess: boolean;
     @observable private _userEmail?: string;
+    @observable private _loginInProgress: boolean;
 
     constructor() {
         this.clear();
+        this._loginInProgress = true;
```
kronken: Move loginInProgess to app state

src/util/ApiClient.ts
```diff
@@ -38,6 +38,11 @@ class ApiClient {
             RequestMethod.POST, endpoints.AUTH, requestBody);
     }
 
+    public postRequest = async (endpoint: endpoints, object: any) => {
```
kronken: object as optional?

src/components/App.tsx
Divinare: Changing it to use getRequest instead

Divinare:
Review fixes done

PR #695, opened 2019-03-11
689 fix bug node moving only horizontal

### LINKED ISSUE #689: Fix dragging node only updates location horizontally.
Seems to be caused by race condition from InputContainer when both lon and lat values change at the same time.

### LINKED ISSUE #690: Manual NodeCoordinates inputs should use updateNodeGeometry instead of updateNode method.

kronken:
- Fixing bug that prevented nodes to be moved vertically
- Rounding geometries modified by node view
- Setting node geometries onChange to use updateGeometries() instead of updateNode()

Closes #689 
Closes #690


Divinare:
Hmm, if I try to open nodeStore with react debugger, it throws an error:
```
nodeStore.ts:38 Uncaught TypeError: Cannot read property 'coordinatesProjection' of undefined
    at Object.get (nodeStore.ts:38)
```

No idea when this came from but it could be fixed in this task. Also now rounding is done in nodeCoordinatesView but I guess its a wrong place. Should be rounded at nodeStore in updateNodeGeometry method.

(because if someone wants to know where coordinate gets rounded after dragging & releasing on map, its strange that form's input view rounds it)

kronken:
Moved the round logic to node store.

Divinare:
React console errors will be fixed here: #706

Issue #662, opened 2019-02-27
Add error param to errorStore's addError


Divinare:
* Also: should try catch be done inside services?

Make errorStore to print error + stack trace to the console
```
    @action
    public addMessage(message: string, error?: Error) {
        this._errors.push(message);
        /* tslint:disable no-console */
        if(error) console.error(error);
        /* tslint:enable no-console */
    }
```

* Also search for ".catch(", we now have catch (ex) and catch (err) --> pick 1 for consistency

PR #705, opened 2019-03-12
619 Inform user that the local db is out of sync

### LINKED ISSUE #619: Show error message to user if jore-map-ui DB was out of sync with Jore DB (when updating node or routePath)
* Update function in our DB throws 409 error code. If frontend receives 409, show error "Karttakäyttöliittymän sisäinen tietokanta oli epäsynkassa jore tietokannan kanssa. Päivitä sivu ja yritä uudelleen."

kronken:
Closes #619 

Test by:
- Make some item out of sync
(Easiest way is to comment node syncing in BE, and then update a nodes name. This save will be successful, but will not be synced to local DB, resulting in a confict)
- Save the item


src/components/sidebar/routePathView/RoutePathView.tsx
```diff
@@ -173,8 +174,7 @@ class RoutePathView extends ViewFormBase<IRoutePathViewProps, IRoutePathViewStat
 
             DialogStore.setFadeMessage('Tallennettu!');
         } catch (err) {
-            const errMessage = err.message ? `, (${err.message})` : '';
-            this.props.errorStore!.addError(`Tallennus epäonnistui${errMessage}`);
+            showError(err);
```
Divinare: showError is too generic name.

src/components/sidebar/networkView/nodeView/NodeView.tsx
```diff
@@ -98,8 +99,7 @@ class NodeView extends ViewFormBase<INodeViewProps, INodeViewState> {
             this.props.nodeStore!.setCurrentStateAsOld();
             this.props.dialogStore!.setFadeMessage('Tallennettu!');
         } catch (err) {
-            const errMessage = err.message ? `, (${err.message})` : '';
-            this.props.errorStore!.addError(`Tallennus epäonnistui${errMessage}`);
+            showError(err);
```
Divinare: What was wrong with how it was before?

Divinare: As in #662, I would add error param to addMessage

```
this.props.errorStore!.addError(Tallennus epäonnistui., error);

@action
public addMessage(message: string, error?: Error) {
    let msg = message;
    const statusCode = error.errorCode;
    if(error.errorCode) {
           msg += ' ' + httpStatusCodeList[error.errorCode];
     }
    this._errors.push(msg);
    /* tslint:disable no-console */
    if(error) console.error(error);
    /* tslint:enable no-console */
}
```

kronken: Refactor to using logic more similar to this example

PR #724, opened 2019-03-15
Select node & link tool implemented

### LINKED ISSUE #602: Network editing: Select link / node tool
When no tool is selected, this tool could be on by default

* clicking link opens linkView where user can edit that link
* clicking a node opens NodeView where user can edit 3 node locations

Divinare:
Closes #602

* Network links are now slightly wider so that they would be easier to click
* NodeView didn't need that componentDidUpdate to refresh component which is strange, linkView however needed it


Divinare:
This can be merged now

PR #726, opened 2019-03-18
601 remove strings to dropdown

### LINKED ISSUE #601: Remove ability to give string list to dropdown.

kronken:
Closes #601 

Also added comment to maybe remove RouteLinkView

### DIFF
--- a/src/components/controls/Dropdown.tsx
+++ b/src/components/controls/Dropdown.tsx
@@ -14,7 +14,7 @@ interface IDropdownBaseProps {
 }
 
 interface IDropdownProps extends IDropdownBaseProps {
-    items: string[] | IDropdownItem[];
+    items: IDropdownItem[];
 }
 
 interface IDropdownWithCodeListProps extends IDropdownBaseProps {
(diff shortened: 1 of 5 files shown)


src/components/controls/Dropdown.tsx
```diff
@@ -8,13 +8,14 @@ export interface IDropdownItem {
 
 interface IDropdownBaseProps {
     label?: string;
-    selected: string;
+    selected?: string;
     disabled?: boolean;
+    allowEmpty?: boolean;
```
Divinare: isEmptyAllowed

src/components/controls/Dropdown.tsx
```diff
@@ -30,28 +31,22 @@ const usesCodeList = (
...
-<IDropdownProps | IDropdownWithCodeListProps> {
+class Dropdown extends React.Component<IDropdownProps | IDropdownWithCodeListProps> {
     onChange = (event: any) => {
         this.props.onChange(event.target.value);
     }
 
     public render() {
-        let dropDownItemList: IDropdownItem[];
+        let dropDownItemList: IDropdownItem[] =
+            this.props.allowEmpty ? [{ key: '', value: '-' }] : [];
```
Divinare: Is this the correct way to implement? How do you know what empty value is? What if empty value would be '' or null or undefined etc and not '-'?

What if you give emptyValue?: any: as props, if that is given, that will be the empty value and it will be used.

Divinare: Also that empty value should be used here I think:
{Boolean(selectedItem) ? selectedItem!.value : ''}

kronken: Good comment, this was refactored now

PR #746, opened 2019-03-20
603 create link tool

### LINKED ISSUE #603: CreateLinkTool
Backend PR: Backend PR: https://gitlab.hsl.fi/jore/jore-map-backend/merge_requests/35

CreateLinkTool:
* switch on networkNodes
phase 1: select startNode, select endNode
*** a new link is drawn between startNode and endNode ***
phase 2: start EditLinkTool

Divinare:
Closes #603 

Backend PR: https://gitlab.hsl.fi/jore/jore-map-backend/merge_requests/35

* DialogStore's setFadeMessage now returns a promise so that redirect to /link/... can be made after 'Tallennettu' text disappears

* Added comment to navigator "prevent new pushing url if the current rl is already the right one"
    * this happened if url was already new/link and user created a new link

* sometimes when creating new link it didn't work at the first attempt. Couldn't reproduce this error anymore though.

* Didn't want to refactor sending whole node when saving ILink
   * added 2 TODOs: "TODO: refactor as startNodeId? TODO: refactor as endNodeId?"

New tasks:

* Save button is now disabled if link from selected startNode --> endNode exist with the current transitType. There should be some visualization for the user of what is wrong, card -> #747 

* Create link / node: fill in address information automatically. Card -> #736 

Test instructions:

* Test that you can't create links with existing transitType + start & end node
   * This is not however prevented in backend -> Save should throw error if foreign constraint is broken
   * Added a test card for link creation: #749

* Test that back button works after link creation

* Test updating existing link

etc...


Divinare:
Review fixes done:

* App breaks when creating two links each other: fixed in this PR. Now new link url is link/new/startNodeId,endNodeId
   * also, user has to select transitType before saving
* Allow creating link beginning where current link ends: #762
* Re-renders network many times when adding link; #754

src/stores/linkStore.ts
```diff
@@ -75,11 +97,22 @@ export class LinkStore {
         };
     }
 
+    @action
+    public setStartMarkerCoordinates = (startMarkerCoordinates: LatLng | null) => {
+        this._startMarkerCoordinates = startMarkerCoordinates;
+    }
+
+    @action
+    public setExistingTransitTypes = (usedTransitTypes: TransitType[]) => {
```
kronken: This variable is probably not required to be in the store?
Actually this variable could be equally well a normal local variable in the component. Since, there is no need for this variable to be observable. We just set and read it.

Divinare:
Review fixes done

src/components/sidebar/linkView/LinkView.tsx
```diff
@@ -127,16 +129,11 @@ class LinkView extends ViewFormBase<ILinkViewProps, ILinkViewState> {
...
 
         const link = this.props.linkStore!.link;
-        if (!link || !link.geometry || !link.startNode || !link.endNode) {
-            navigator.goTo(SubSites.home);
-            return;
-        }
-
         const existingLinks = await LinkService.fetchLinks(link.startNode.id, link.endNode.id);
         if (existingLinks.length > 0) {
             const existingTransitTypes = existingLinks
```
kronken: this.existingTransitTypes =

PR #755, opened 2019-03-21
754 bug rendering of network

### LINKED ISSUE #754: Fix re-rendering of node/link/linkPoint layers
* One way is to add <VectorGridLayer key += 1 />

kronken:
- Prevent stacking of reaction calls
- Prevent recreating of node and link when they are updated

Closes #754


src/components/map/layers/NetworkLayers.tsx
```diff
@@ -236,6 +238,11 @@ class NetworkLayers extends Component<INetworkLayersProps> {
         return { className: s.hidden };
     }
 
+    private recreateReaction = (type: GeoserverLayer) => (reaction: IReactionDisposer) => {
```
Divinare: setVectorgridLayerReaction?

I would add a comment too:
// Sets a reaction object for GeoserverLayer (replaces existing one) so that reaction object's wouldn't multiply each time a VectorGridLayer is re-rendered.

kronken:
Fixed.

I also added observer() wrapper to all functional components, which made me notice that we had several very simple react classes that should be made as functions. I made a simple refactor, but due to this, the change log is a bit messy.
In order to easier review this, the only files that have functional changes are:
nodeStore
linkStore
VectorGridLayer
NetworkLayers
All other files are just refactoring.

PR #759, opened 2019-03-22
715 Routepath validations in BE

### LINKED ISSUE #715: RoutePath validations backend
* validate: const routePath = req.body as IRoutePath;

* routePath's start & end times should have clock 04:00 (research)
FE:
https://github.com/HSLdevcom/jore-map-ui/pull/759
BE: 
https://gitlab.hsl.fi/jore/jore-map-backend/merge_requests/37

kronken:
Closes #715


src/factories/routePathFactory.ts
```diff
@@ -49,6 +49,9 @@ class RoutePathFactory {
     }
 
     public static createNewRoutePath(lineId: string, route: IRoute): IRoutePath {
+        const defaultDate = new Date();
+        defaultDate.setHours(0, 0, 0, 0);
```
Divinare: Move to constants? Something like:
NEW_ROUTE_PATH_HOURS: 0

kronken: Standard dates are always written as midnight, so I would assume this will never change, and therefore better be hard coded.

PR #761, opened 2019-03-22
662 error param to error store

### LINKED ISSUE #662: Add error param to errorStore's addError

kronken:
Closes #662

### DIFF
--- a/src/components/map/tools/ExtendRoutePathTool.ts
+++ b/src/components/map/tools/ExtendRoutePathTool.ts
@@ -1,7 +1,6 @@
 import RoutePathStore from '~/stores/routePathStore';
 import NodeType from '~/enums/nodeType';
 import ToolbarTool from '~/enums/toolbarTool';
-import ErrorStore from '~/stores/errorStore';
 import { INode } from '~/models';
 import RoutePathLinkService from '~/services/routePathLinkService';
 import BaseTool from './BaseTool';
@@ -23,25 +22,19 @@ class ExtendRoutePathTool implements BaseTool {
     }
 
     public onNetworkNodeClick = async (clickEvent: any) => {
-        try {
-            if (!this.isNetworkNodesInteractive()) return;
+        if (!this.isNetworkNodesInteractive()) return;
 
-            const properties = clickEvent.sourceTarget.properties;
-            if (properties.soltyyppi !== NodeType.STOP) return;
-            const queryResult =
-                await RoutePathLinkService.fetchNeighborRoutePathLinks(
-                    properties.soltunnus,
-                    1,
-                    RoutePathStore!.routePath!.transitType,
-                    RoutePathStore!.routePath!.routePathLinks,
-                );
-            RoutePathStore!.setNeighborRoutePathLinks(queryResult!.routePathLinks);
-            RoutePathStore!.setNeighborToAddType(queryResult!.neighborToAddType);
-
-        } catch (ex) {
-            ErrorStore.addError(
-                'Unable to process network node click using ExtendRoutePathTool', ex);
-        }
+        const properties = clickEvent.sourceTarget.properties;
+        if (properties.soltyyppi !== NodeType.STOP) return;
+        const queryResult =
+            await RoutePathLinkService.fetchNeighborRoutePathLinks(
+                properties.soltunnus,
+                1,
+                RoutePathStore!.routePath!.transitType,
+                RoutePathStore!.routePath!.routePathLinks,
+            );
+        RoutePathStore!.setNeighborRoutePathLinks(queryResult!.routePathLinks);
+        RoutePathStore!.setNeighborToAddType(queryResult!.neighborToAddType);
     }
 
     public onNodeClick = (node: INode, linkOrderNumber: number) => async () => {
--- a/src/components/map/tools/ExtendRoutePathTool.ts
+++ b/src/components/map/tools/ExtendRoutePathTool.ts
@@ -39,7 +39,8 @@ class ExtendRoutePathTool implements BaseTool {
             RoutePathStore!.setNeighborToAddType(queryResult!.neighborToAddType);
 
         } catch (ex) {
-            ErrorStore.addError((ex as Error).message);
+            ErrorStore.addError(
+                'Unable to process network node click using ExtendRoutePathTool', ex);
         }
     }
 
--- a/src/models/IError.ts
+++ b/src/models/IError.ts
@@ -1,0 +1,3 @@
+export default interface IError extends Error {
+    errorCode?: number;
+}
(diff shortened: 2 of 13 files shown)


src/models/IError.ts
```diff
@@ -0,0 +1,3 @@
+export default interface IError extends Error {
+    errorCode?: number;
```
Divinare: Forgot to add these message & name properties?

src/components/map/tools/ExtendRoutePathTool.ts
```diff
@@ -39,7 +39,8 @@ class ExtendRoutePathTool implements BaseTool {
             RoutePathStore!.setNeighborToAddType(queryResult!.neighborToAddType);
 
         } catch (ex) {
-            ErrorStore.addError((ex as Error).message);
+            ErrorStore.addError(
+                'Unable to process network node click using ExtendRoutePathTool', ex);
```
Divinare: Use finnish

src/stores/errorStore.ts
```diff
@@ -22,14 +23,14 @@ export class ErrorStore {
     }
 
     @action
-    public addError(message: string, error?: Error) {
+    public addError(message: string, error?: IError) {
```
Divinare: Hmm, error is always given as param. I would change: error? -> error, error as required param

src/util/ApiClient.ts
```diff
@@ -73,12 +69,13 @@ class ApiClient {
...
             }
             error = {
+                name: 'Failed to fetch',
                 errorCode: response.status,
                 message: response.statusText,
             };
         } catch (err) {
-            console.log(err); // tslint:disable-line no-console
             error = {
+                name: 'Connectivity error',
```
Divinare: Name field is only being used within tests (if i'm not blind) so why add name property? In tests, you can just compare error message

src/stores/__tests__/errorStore.test.ts
```diff
@@ -0,0 +1,83 @@
...
+
+        expect(errorStore.errorCount).toEqual(1);
+
+        const error = errorStore.pop();
+
+        expect(error).toEqual(errorMessage);
+        expect(errorStore.errorCount).toEqual(0);
+    });
+
+    it('Adds one error with exception to error store', () => {
```
Divinare: I would change name of this test as "Adds one error to error store" if error is made as required param

src/stores/__tests__/errorStore.test.ts
```diff
@@ -0,0 +1,83 @@
...
+        // tslint:disable:no-console
+        // Mocks console.error
+        const consoleError = console.error;
+        global.console.error = jest.fn();
+
+        const errorStore = new ErrorStore;
+        const errorMessage = 'This is a test error';
+        const exceptionMessage = 'This is a test exception message';
```
Divinare: Hmm, exception & error... both mean the same. errorMessage1, errorMessage2? I wouldn't use exception word

See my comment @ this card's description:
Also search for ".catch(", we now have catch (ex) and catch (err) --> pick 1 for consistency

There is still one catch(err) ...

I would use error everywhere (addError, errorMessage, catch(err), etc)

src/models/IError.ts
kronken: IError extends Error, message and name properties are included in Error interface

src/components/map/tools/ExtendRoutePathTool.ts
kronken: oh

kronken: true dat

src/stores/errorStore.ts
kronken: Hmm, really? That would remove the possibility for us to create simple error notification.
Instead we would always need to come up with an exception.

src/util/ApiClient.ts
kronken: Name is a default property of the Typescript Error interface

src/stores/__tests__/errorStore.test.ts
kronken: Error does not mean the same thing as an exception. Error means a generic issue that has come up. We could basically write 'Lost contact with the server', without ever throwing an exception. Exception is a standard for throwing and catching errors.

src/stores/errorStore.ts
kronken: Check https://github.com/HSLdevcom/jore-map-ui/blob/20c626455a28815a44765c6a97e11a1cb27c6656/src/services/routePathLinkService.ts#L79

src/stores/__tests__/errorStore.test.ts
kronken: I would not make error a mandatory property, we currently have use cases where we show error messages, where it would be unneccessary to add an exception:
https://github.com/HSLdevcom/jore-map-ui/blob/20c626455a28815a44765c6a97e11a1cb27c6656/src/services/routePathLinkService.ts#L79

src/components/map/tools/ExtendRoutePathTool.ts
kronken: Do you have a available translation for: Unable to process network node click using ExtendRoutePathTool

src/models/IError.ts
Divinare: Ah I see

src/components/map/tools/ExtendRoutePathTool.ts
Divinare: Perhaps `Solmusta (soltunnus: ${properties.soltunnus} alkavien linkkien haku epäonnistui`

But hmm, seems like routePathLinkServie already does a try catch... maybe we should remove it if we decided to try catch outside of service

src/stores/__tests__/errorStore.test.ts
Divinare: https://stackoverflow.com/questions/16142583/whats-the-difference-between-error-and-exception-in-javascript

says: Behind the scenes, the browsers use the same window.Error constructor. An Exception is an Error instance with a name and message that contain "Exception"

So okay that is an exception if it has name & message fields.

```
        const errorStore = new ErrorStore;
        const errorMessage = 'This is a test error';
        const exception: Error = {
            message: 'This is a test exception message',
            name: 'Test exception',
        };
        errorStore.addError(errorMessage, exception);
```

Divinare: Based on that link, we should maybe use catch (e) because we don't know if its an exception or error

PR #769, opened 2019-03-27
707 Logout on forbidden

### LINKED ISSUE #707: Prevent user from having her session expired while using the application
One solution would be to have the session expiration time to roughly 4 hours, and update this value every time the backend gets a call.

- Current implementation seems to do this automatically
https://gitlab.hsl.fi/jore/jore-map-backend/merge_requests/36

FE:
https://github.com/HSLdevcom/jore-map-ui/pull/769

BE:
https://gitlab.hsl.fi/jore/jore-map-backend/merge_requests/36

kronken:
Closes #707 

Automatically logs out the user when receiving a Forbidden response 403


Divinare:
Bug: logout doesn't work.

Test with changing backend's maxAge to 5000 and loginRequired constants true
--> getting error

Divinare:
Not working: if you keep clicking, backend still logs you out

PR #775, opened 2019-03-27
710 Automatically set editing when geometry change

### LINKED ISSUE #710: After editing geometry,  save should be possible without clicking edit pen
* clicking toolbar icon activates edit pen

kronken:
Closes #710 

Edit RoutePath, Link or Node geometry. Edit mode should automatically be turned on.


src/components/sidebar/linkView/LinkView.tsx
```diff
@@ -65,6 +65,7 @@ class LinkView extends ViewFormBase<ILinkViewProps, ILinkViewState> {
     }
 
     async componentDidMount() {
+        super.componentDidMount();
```
Divinare: Can these calls be left out? Or is there any other better way?

kronken: Without these, the super functions will never be called. So, this is the most elegant way I found.

Divinare:
Testing: clicking edit pen when its active shows a warning "Tallentamattomat muutokset kumotaan" but it doesn't reset undoStore. So undoing does change geometry which puts edit pen back to active

I think unactivating edit pen should also reset undoStore

Found also an interesting situation after unactivating edit pen:
afterUnactivatedEdit

Issue #737, opened 2019-03-19
Make routePath's node/link list filters also filter what is shown on map (?)


Divinare:
* Is this really wanted? I mean, it would be pretty ugly to show only nodes on map

PR #780, opened 2019-03-28
725 remove route path link view

### LINKED ISSUE #725: RoutePath link / node click should open item in routePath list, remove RoutePathLink view

Divinare:
Closes #725 

* commented out routePath infoTab's open link and throwing alert "Suunnittelu kesken" (before it redirected to routePathLink view)
   * I would say we should remove that table completely so I didn't fix that functionality

* showItem is now a single param (for both links & nodes). Added a mention to #737 to maybe split them into showNode & showLink

* listFilters moved to routePathStore so that routePathView can change them

* removed linkBuilder

* Added planning card: #781


src/components/sidebar/routePathView/RoutePathView.tsx
```diff
@@ -1,13 +1,13 @@
 import React from 'react';
-import moment from 'moment';
+import Moment from 'moment';
```
kronken: moment is a function, the same as classnames is a function, that we import directly from a dependency. It is not a class, like the React class we import from react. Functions start with lowercase letter, Classes start with uppercase.

Divinare: Fixes done, went with Moment (its namespace)

Divinare: This uses camel case: https://www.typescriptlang.org/docs/handbook/namespaces.html

PR #786, opened 2019-04-01
search address field implemented

### LINKED ISSUE #733: Input field for searching an address and jumping to it on map
* https://github.com/HSLdevcom/hsl-route-map-poc/blob/master/index.html#L304

https://api.digitransit.fi/geocoding/v1/search?text=pasila,helsinki

* show label as a search result

http://kartat.hsl.fi/linjakartta/

Divinare:
Closes #733

* Added TODO to move createDivIcon to /utils. Wasn't sure because maybe it doesn't belong there and can as well be where it is (or could be moved to some other place)
* Had to modify ApiClient a bit so that we could send request without 'include' param

### DIFF
--- a/src/components/map/AddressSearch.tsx
+++ b/src/components/map/AddressSearch.tsx
@@ -70,6 +70,8 @@ class AddressSearch extends Component<IAddressSearchProps, IAddressSearchState> 
             this.unselectSearchResult();
             return;
         }
+        if (this.searchResultMarker) this.map.removeLayer(this.searchResultMarker);
+
         const coordinates = selectedSearchResult.geometry.coordinates;
         const latLng = L.latLng(coordinates[1], coordinates[0]);
         this.setState({
--- a/src/components/map/AddressSearch.tsx
+++ b/src/components/map/AddressSearch.tsx
@@ -115,11 +115,11 @@ class AddressSearch extends Component<IAddressSearchProps, IAddressSearchState> 
         }
     }
     private requestAddress = async (value: string) => {
-        const GEOCODER_ADDRESS = 'https://api.digitransit.fi/geocoding/v1/search';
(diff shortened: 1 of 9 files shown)


src/components/map/AddressSearch.tsx
```diff
@@ -0,0 +1,194 @@
...
+            this.setState({
+                input: newValue,
+            });
+            this.requestAddress(newValue);
+        } else {
+            this.unselectSearchResult();
+        }
+    }
+    private requestAddress = async (value: string) => {
+        const GEOCODER_ADDRESS = 'https://api.digitransit.fi/geocoding/v1/search';
```
kronken: Address in constants

src/util/ApiClient.ts
```diff
@@ -46,15 +48,26 @@ class ApiClient {
...
     private sendRequest = async (method: RequestMethod, endpoint: endpoints, object: any) => {
+        const entityUrl = this.getEntityUrl(endpoint);
+        return this.sendRequestToUrl(method, entityUrl, object, 'include');
+    }
+
+    private getEntityUrl = (entityName: endpoints) => {
+        return `${API_URL}/${entityName}`;
+    }
+
+    public sendRequestToUrl = async (
```
kronken: the difference in name between sendRequest and sendRequestToUrl is not declarative enough.

kronken: I would suggest:
sendRequestToUrl => sendRequest
sendRequest => sendJoreMapRequest

kronken:
Found bug:
Cannot click on first search result

src/util/ApiClient.ts
Divinare: Fixed, its much better now

PR #800, opened 2019-04-02
721 validation bud when switching node type

### LINKED ISSUE #721: Switching nodeView's nodeType prevents saving if validations were broken
Steps:
1. Open node that is a stop
1. Change some field so that validation fails
1. Change nodeType to crossroad
--> save button is grey

This seems to be fixed in master, I can't get this problem to occur. NO PR.

kronken:
Closes #721 

Fixes:
- bug that prevents you from saving crossroad/border if underlying stop is invalid.
- bug that does not validate node fields in stop view.
- bug that does not automatically show error messages when changing type from node to stop.


Divinare:
New PR: https://github.com/HSLdevcom/jore-map-ui/pull/809

PR #828, opened 2019-04-17
789 node area codes

kronken:
Test by creating new node.

- Added transit toggle, which we need in order to build id
- Also added functionality for showing error message that comes from BE.


src/util/ApiClient.ts
```diff
@@ -88,10 +88,15 @@ class ApiClient {
                 });
                 return;
             }
+            let message = response.statusText;
+            const errorContent = await response.text();
+            if (errorContent) {
+                message = errorContent;
+            }
             error = {
+                message,
```
Divinare: I would go with:
```
            // error
            const responseText = await response.text();
            const errorMessage = responseText ? responseText : response.statusText;
            error = {
                message: errorMessage,
                name: 'Failed to fetch',
                errorCode: response.status,
            };
```

Divinare:
Testing:

* saving a stop is possible without selecting transitType - why?

* now if local coordinates are not found, user gets an error "Tallennus epäonnistui, Unable to find local area for coordinates: 22.774658,61.043662"
--> is there a card for this? I think that error should be in finnish to indicate that its all fine, location was just not found from our local area service. Error message could be something like "Tallennus epäonnistui, solmun tunnusta ei voitu luoda. Syy: koordinaateille 61.043662, 22.774658 ei löytynyt aluekoodia. Pyydä tarvittaessa ylläpitotiimiltä lisäämään koordinaateille aluekoodi"

bug:
* when nameFi and nameSw are valid, clicking new node opens up form that allows those fields to be saved as empty

src/util/ApiClient.ts
kronken: This is indeed better

kronken:
As answer to your questions:
1: Empty selection currently results in 3 ( = Ei varattu)
2: https://github.com/HSLdevcom/jore-map-ui/issues/835
3: https://github.com/HSLdevcom/jore-map-ui/issues/827

PR #836, opened 2019-04-24
829 fix login bug

### LINKED ISSUE #829: Login says at first login "Tämä toimenpide on vanhentunut, ohjaamme sinut takaisin kirjautumissivulle"
I guess this is because of token getting old. Test if you can reproduce the bug with making token time 5 sec and logging in after a while.

Divinare:
Closes #829 

* test with both isLodingRequired true/false in BE / FE. Also put 5000 ms into server.ts cookie max age.

* initCodeList call is now in a bit odd place but wasn't able to get it work in init() func. But I think it should be fine where it is now

### DIFF
--- a/src/components/App.tsx
+++ b/src/components/App.tsx
@@ -38,22 +38,26 @@ interface IAppProps extends RouteComponentProps<any> {
 @observer
 class App extends React.Component<IAppProps, IAppState> {
 
-    private renderApp = observer(() => (
-        <>
-            <NavigationBar />
-            <div className={s.appContent}>
-                <div className={this.props.mapStore!.isMapFullscreen ? s.hidden : ''}>
-                    <Sidebar
-                        location={this.props.location}
-                    />
+    private renderApp = observer(() => {
+        this.initCodeLists();
+
+        return (
+            <>
+                <NavigationBar />
+                <div className={s.appContent}>
+                    <div className={this.props.mapStore!.isMapFullscreen ? s.hidden : ''}>
+                        <Sidebar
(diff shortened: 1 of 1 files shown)


src/components/App.tsx
```diff
@@ -64,11 +68,7 @@ class App extends React.Component<IAppProps, IAppState> {
     }
 
     componentWillMount() {
```
kronken: componentWillMount is legacy, and shouldn't really be used.

src/components/App.tsx
```diff
@@ -80,20 +80,22 @@ class App extends React.Component<IAppProps, IAppState> {
         }
     }
 
-    private redirectToLogin = async () => {
+    private init = async () => {
```
kronken: Overall I find that the logic could be improved. It feels like a race condition that we call async method init, without awaiting it. What happens if init call takes longer than what it takes on our development machines. If init takes longer, then application will load as usually, but there will be errors, since we are not logged in. Then finally when the init() method completes, it will redirect the user to login page. I think this should be made so that the page is viewed right from the start, but App.tsx should only show a loading icon, as long as init() function is loading. When init function is done, we can then let the user access the app, or redirect the user to login page.

kronken: I know that your PR do not have anything to do with this thing, we could probably create an issue to fix this.

Divinare:
Changed componentWillMount as componentDidMount. Also some minor refactoring + changed "Logging in... into Kirjaudutaan sisään"

PR #901, opened 2019-05-16
fixed inputContainer height, removed textContainer from inputContainer

### LINKED ISSUE #899: Make text field height as some static value when disabled

Divinare:
Closes #899


src/components/controls/TextContainer.tsx
```diff
@@ -7,6 +7,7 @@ interface IInputProps {
     label: string | JSX.Element;
     value?: string | number | null | Date;
     darkerInputLabel?: boolean;
+    disabled?: boolean;
```
kronken: Why does a TextContainer that is never editable, a prop named disabled?

Divinare: Now TextContainer has always static height

Issue #895, opened 2019-05-15
Hide modifiedOn and ModifiedBy if the entity is new


kronken:
Also fix timestamps to be shown with correct format

Make modifiedOn as optional in each model, parse nodifiedOn as a date in each factory.

Is this needed? T. Joe

PR #903, opened 2019-05-17
Displaying login error properly

### LINKED ISSUE #663: Get username from jr_kayttaja table and use that when saving
FE PR: https://github.com/HSLdevcom/jore-map-ui/pull/903
BE PR: https://gitlab.hsl.fi/jore/jore-map-backend/merge_requests/67

When signing in and user has write access, backend tries to get user with hsl-id account's email address
* if email is not found, redirect user to login page and show error to the user "Kirjautuminen epäonnistui, sähköpostiosoitteen omaavaa käyttäjätunnusta ei löytynyt joukkoliikennerekisterin tietokannasta. Ota yhteyttä käyttäjätunnusten hallinnoijaan."

Divinare:
Closes #663 

BE PR: https://gitlab.hsl.fi/jore/jore-map-backend/merge_requests/67

- textContainer fails to display "MUOKATTU" timestamp, edited description of card #895 to fix it
- showing login failure information if username was not found
   - also logging out when login fails


src/util/text.ts
```diff
@@ -0,0 +1,19 @@
+import textCodeList from '~/codeLists/textCodeList';
```
kronken: The naming needs to be changed. 
A file named text.ts, with a function named text doesn't tell that much

src/util/text.ts
```diff
@@ -0,0 +1,19 @@
...
+
+/**
+ * @param {keyValueMap} { key: value } - key is the same as ${key} in textCodeListKey, ${key} is replaced with value
+ **/
+const text = (textCodeListKey: string, keyValueMap?: Object) => {
+    let s = textCodeList[textCodeListKey];
+    const re = /\$\{(\w+)\}/g; // ${...}
```
kronken: const regexRule

src/util/text.ts
```diff
@@ -0,0 +1,19 @@
+import textCodeList from '~/codeLists/textCodeList';
+
+/**
+ * @param {keyValueMap} { key: value } - key is the same as ${key} in textCodeListKey, ${key} is replaced with value
+ **/
+const text = (textCodeListKey: string, keyValueMap?: Object) => {
+    let s = textCodeList[textCodeListKey];
```
kronken: let message

Divinare: I put lineString there now

src/util/text.ts
Divinare: CodeListHelper & getText

Divinare:
Review fixes ^

src/util/text.ts
kronken: sure, everything is better than 's' :D

PR #955, opened 2019-07-05
951 reimplement dropdown with search

### LINKED ISSUE #951: Implement dropdown with search input
* Use in stop area dropdown
* Filling input field filtters options
* Value is correct only if it exists the option list
   * how to indicate invalid value: show text "Valitse pysäkkialue alasvetovalikosta"

Divinare:
Closes #951 

Dropdown now reimplemented. Using react-select library
* to fix lag issue with search results being > 1000, results are cut down to max 200, also own filter handler was needed
* all other dropdowns have now new styling

### DIFF
--- a/package.json
+++ b/package.json
@@ -52,6 +52,7 @@
         "react-moment": "^0.7.9",
         "react-router": "^4.3.1",
         "react-router-dom": "^4.3.1",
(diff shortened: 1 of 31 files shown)


Divinare:
Self review & test ok

PR #959, opened 2019-08-14
lineTopic form

### LINKED ISSUE #956: LineHeader form
FE PR: https://github.com/HSLdevcom/jore-map-ui/pull/959
BE PR: https://gitlab.hsl.fi/jore/jore-map-backend/merge_requests/85

Copy the idea from the old UI:
image

Divinare:
#closes #956 

BE PR: https://gitlab.hsl.fi/jore/jore-map-backend/merge_requests/85

* edit / save a new lineHeader
* note that even startDate of a lineHeader can be edited (startDate is part of LineTopic primarykey which is lineId + startDate combination)
   * this is why originalStartDate is being sent to backend when saving


src/components/sidebar/lineView/LineInfoTab.tsx
```diff
@@ -48,16 +49,16 @@ class LineInfoTab extends React.Component<
...
     }
 
     async componentWillMount() {
         const lineId = this.props.lineStore!.line!.id;
-        const lineTopics: ILineTopic[] = await LineTopicService.fetchLineTopicsForLineId(
+        const lineHeaders: ILineHeader[] = await LineHeaderService.fetchLineHeaders(
             lineId
         );
-        this.initLineTopicItems(lineTopics);
+        this.initLineHeaderItems(lineHeaders);
```
ahjyrkia: Maybe unnecessary to make this in to a function of its own? Just moving around lineHeaders and not making changes to it.

src/components/sidebar/lineView/LineHeaderTable.tsx
```diff
@@ -60,42 +72,41 @@ class LineTopicTable extends React.Component<ILineTopicListProps> {
                                 </th>
                                 <th className={s.columnHeader} />
                             </tr>
-                            {this.props.lineTopics.map(
-                                (lineTopic: ILineTopic, index: number) => {
-                                    const isCurrentLineTopic = _.isEqual(
-                                        this.props.currentLineTopic,
-                                        lineTopic
+                            {this.props.lineHeaders.map(
```
ahjyrkia: Maybe move this to own function to make render-function easier to read. Looks a bit messy right now but the reason for this might also be the current row length limit.

{ this.lineHeaderRows() }

src/components/sidebar/lineView/LineHeaderTable.tsx
```diff
@@ -4,34 +4,46 @@ import _ from 'lodash';
...
+            .toLink();
+
+        navigator.goTo(newLineHeaderLink);
     };
     render() {
         return (
-            <div className={s.lineTopicTableView}>
+            <div className={s.lineHeaderTableView}>
                 <SidebarHeader hideCloseButton={true}>
                     Linjan otsikot
```
ahjyrkia: Not part of this pr but why is the title indented?

ahjyrkia: Not part of this pr but why is the title indented?

src/components/sidebar/lineView/LineHeaderTable.tsx
```diff
@@ -60,42 +72,41 @@ class LineTopicTable extends React.Component<ILineTopicListProps> {
...
+                                        this.props.currentLineHeader,
+                                        lineHeader
                                     );
                                     return (
                                         <tr
                                             key={index}
                                             className={
-                                                isCurrentLineTopic
-                                                    ? s.lineTopicRowHighlight
+                                                isCurrentLineHeader
```
ahjyrkia: Use classnames? 
```
className={classnames(
  s.lineHeaderRow,
  isCurrentLineHeader ? s.lineHeaderRowHighlight : null
)}
```

src/components/sidebar/lineView/LineInfoTab.tsx
Divinare: We try to keep react's lifecycle methods as clean as possible. The details of that code is not very interesting so it is better to put it in its own method

src/components/sidebar/lineView/LineHeaderTable.tsx
Divinare: Prettier indents it automatically. We need to change that 100 char per line restriction at some point

src/components/sidebar/lineView/LineHeaderTable.tsx
Divinare: There is no such class as .lineHeaderRow?

PR #974, opened 2019-09-09
WIP: initial commit for kilpiVia input field and editing

ahjyrkia:
closes https://github.com/HSLdevcom/jore-map-ui/issues/874

BE PR: https://gitlab.hsl.fi/jore/jore-map-backend/merge_requests/95#38c10652728178d35b9d04461227eaef0df8cd77

### DIFF
--- a/src/components/sidebar/routePathView/RoutePathView.tsx
+++ b/src/components/sidebar/routePathView/RoutePathView.tsx
@@ -19,13 +19,14 @@ import { ToolbarStore } from '~/stores/toolbarStore';
 import { ErrorStore } from '~/stores/errorStore';
 import navigator from '~/routing/navigator';
 import ViewFormBase from '~/components/shared/inheritedComponents/ViewFormBase';
-import { IRoutePath } from '~/models';
+import { IRoutePath, IKilpiVia, IRoutePathLink } from '~/models';
 import routePathValidationModel from '~/models/validationModels/routePathValidationModel';
 import RouteService from '~/services/routeService';
 import routeBuilder from '~/routing/routeBuilder';
 import QueryParams from '~/routing/queryParams';
 import SubSites from '~/routing/subSites';
 import RoutePathService from '~/services/routePathService';
+import KilpiViaService from '~/services/kilpiViaService';
 import LineService from '~/services/lineService';
 import ToolbarTool from '~/enums/toolbarTool';
 import EventManager from '~/util/EventManager';
@@ -54,6 +55,7 @@ interface IRoutePathViewState {
     isLoading: boolean;
     invalidPropertiesMap: object;
     isEditingDisabled: boolean;
+    selectedKilpiViaNames: IKilpiVia[];
 }
 
 @inject(
@@ -74,7 +76,8 @@ class RoutePathView extends ViewFormBase<
         this.state = {
             isLoading: true,
             invalidPropertiesMap: {},
-            isEditingDisabled: !props.isNewRoutePath
+            isEditingDisabled: !props.isNewRoutePath,
+            selectedKilpiViaNames: []
         };
     }
 
@@ -176,6 +179,7 @@ class RoutePathView extends ViewFormBase<
     };
 
     private fetchRoutePath = async () => {
+        this.setState({ isLoading: true });
         const [
             routeId,
             startTimeString,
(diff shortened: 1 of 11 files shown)


src/components/sidebar/routePathView/RoutePathView.tsx
```diff
@@ -194,6 +198,36 @@ class RoutePathView extends ViewFormBase<
...
+            const createPromise = async () => {
+              try {
+                const kilpiViaName: IKilpiVia | null = await KilpiViaService.fetchKilpiViaName(routePathLink.id);
+                if (kilpiViaName) kilpiViaNames.push(kilpiViaName);
+              } catch (err) {
+                this.props.errorStore!.addError(
+                    'KilpiVia haku ei onnistunut.',
+                    err
+                )
+              }
```
Divinare: Fix indent, should be 4. Here it seems to be randomly 2-4. Strongly recommend installing prettier and getting it to work when saving (we already have it configured for the project if you look at package.json)

src/components/sidebar/routePathView/RoutePathView.tsx
```diff
@@ -74,7 +76,8 @@ class RoutePathView extends ViewFormBase<
         this.state = {
             isLoading: true,
             invalidPropertiesMap: {},
-            isEditingDisabled: !props.isNewRoutePath
+            isEditingDisabled: !props.isNewRoutePath,
+            selectedKilpiViaNames: []
```
Divinare: Is this needed?

src/components/sidebar/routePathView/routePathListTab/RoutePathListNode.tsx
```diff
@@ -178,6 +179,25 @@ class RoutePathListNode extends ViewFormBase<
...
+        kilpiViaName = {
+          relid: routePathLinkId,
+          viasuomi: '',
+          viaruotsi: ''
+        }
+      }
+      if (language === 'fi') {
+        kilpiViaName.viasuomi = value;
+      } else {
+        kilpiViaName.viaruotsi = value;
```
Divinare: Indent 4

src/components/sidebar/routePathView/routePathListTab/RoutePathListNode.tsx
```diff
@@ -307,6 +329,26 @@ class RoutePathListNode extends ViewFormBase<
                         darkerInputLabel={true}
                     />
                 </div>
+                <div className={s.flexInnerRow}>
+                    <InputContainer
+                        label='VIA KILPI SUOMEKSI'
+                        disabled={isEditingDisabled}
+                        value={kilpiViaName ? kilpiViaName.viasuomi : ''}
+                        validationResult={
+                            invalidPropertiesMap['kilpiViaName']
```
Divinare: Validation model missing, create one. models/validationModels/kilpiViaValidationModel.ts

src/enums/endpoints.ts
```diff
@@ -8,7 +8,8 @@ enum endpoints { // TODO: rename as endpointPath
     ROUTEPATH = 'routePath',
     ROUTELINK = 'routeLink',
     NODE = 'node',
-    LINK = 'link'
+    LINK = 'link',
+    KILPI_VIA = 'kilpivia'
```
Divinare: If having KILPI_VIA on its own endpoint, there will be a problem: if routePath saving fails, also kilpiVia save should be rolled back and vise versa. That one will be hard / unpractical to implement with separate API calls

I think you should save kilpiVia's with same API call routePath so that you can make those savings depend on each other

(also kilpi_via saving is strongly related to saving routePath so thats why it also belongs to the same API call)

src/services/graphqlQueries.ts
```diff
@@ -259,6 +259,32 @@ const getAllLineHeadersQuery = () => {
...
+        kilpiVia: viaKilpiNimetByRelid(relid: $relid) {
+          relid
+          viasuomi
+          viaruotsi
+        }
+      }
+  `;
+};
+
+const getAllViaKilpiNimets = () => {
```
Divinare: getAllKilpiViaNames

src/stores/routePathStore.ts
```diff
@@ -90,11 +100,51 @@ export class RoutePathStore {
         return this._invalidLinkOrderNumbers;
     }
 
+    @computed
+    get kilpiViaNamesHash(): IKilpiViaHash {
```
Divinare: How about:
get kilpiViaName(relid): null | IKilpiViaName { }

There is no need to return the hash, other components doesn't have to know that about that kilpiVia's are stored in a hash

src/stores/routePathStore.ts
```diff
@@ -90,11 +100,51 @@ export class RoutePathStore {
...
+        const kilpiViaNames: IKilpiVia[] = [];
+        const kilpiViaNamesHash = this._kilpiViaNamesHash;
+        for (const k in kilpiViaNamesHash) {
+          kilpiViaNames.push(kilpiViaNamesHash[k]);
+        }
+        return kilpiViaNames;
+    }
+
+    @action
+    public setKilpiViaNamesHash = (kilpiViaNames: IKilpiVia[]) => {
```
Divinare: setKilpiViaNames

src/stores/routePathStore.ts
```diff
@@ -90,11 +100,51 @@ export class RoutePathStore {
...
+      this._kilpiViaNamesHash[kilpiViaName.relid] = kilpiViaName;
+    };
+
+    @action
+    public setOldKilpiViaNames = (kilpiViaNames: IKilpiViaHash) => {
+        this._oldKilpiViaNamesHash = _.cloneDeep(kilpiViaNames);
+    };
+
+    @action
+    public clearKilpiViaNamesHash = () => {
```
Divinare: clearKilpiViaNames

src/models/IKilpiVia.ts
```diff
@@ -0,0 +1,11 @@
+interface IKilpiViaPrimaryKey {
+    relid: number;
+}
+
+export default interface IKilpiVia extends IKilpiViaPrimaryKey {
+    viaruotsi: string,
+    viasuomi: string
```
Divinare: These fields should be named in english like in other models:

```
interface IKilpiViaPrimaryKey {
    id: number;
}
export default interface IKilpiVia extends IKilpiViaPrimaryKey {
    nameSw: string,
    nameFi: string
}
```
Then create
```
IExternalKilpiVia {
    relid: string,
    viaruotsi: string,
    viasuomi: string
}
```

However, since this IKilpiVia is only used at routePathView, i would define interface IExternalKilpiVia inside kilpiViaService and "hide it there". Other components doenst have to know about it. Just map IExternalKilpiVia into IKilpiVia and let that kilpiViaService return it

PR #1020, opened 2019-10-01
916 automatically fill stop postal number and address

Divinare:
* Implemented reverse geocoding using openstreetmap API: https://nominatim.org/release-docs/develop/api/Reverse/
* encountered a problem with current state of ViewFormBase: isEditingDisabled had to be moved to store.
    * also countered other problem: fetching address data in nodeStore -> need to validate stop properties in stop form -> had to implement 3 watch functions to stop form (addressFi, addressSw, postalNumber) to validate those fields changed by address fetch. Couldn't validate those properties in nodeStore (which would have been ideal solution), but instead had to listen changes to validate those properties in stop form because stop form has access to its invalidPropertiesMap
    -> this brought in mind an idea: what if viewFormBase would be in a store that each view would use. This one should be implemented in other card: https://github.com/HSLdevcom/jore-map-ui/issues/1019
* isEditingDisabled had to be moved to nodeStore because it was hard to watch when it changed to fetch address after that (however I could try to revert it back to viewFormBase since codebase changed slightly (address fetch is in nodeStore etc.)
    * refactoring task to move it out from VIewFromBase in other views as well: https://github.com/HSLdevcom/jore-map-ui/issues/1022
---

**Test this task carefully. Try to test every single variation. There are many ways that stop form could break.**

A few bugs fixed here:

Fixed stop form invalidPropertiesMap not reseting:
(test at stage environment)
1) open random stop form
2) change name of a stop to > 20 letters
3) click "isEditingDisabled"
-> error msg stays on screen

Fixed new stop form now validating properly
(test at stage environment)
1) open new stop form with "new stop tool"
2) change name field so that validation passes
3) create a new stop again
-> name field is not validated even though the field is empty

### DIFF
--- a/src/components/map/AddressSearch.tsx
+++ b/src/components/map/AddressSearch.tsx
@@ -135,12 +135,12 @@ class AddressSearch extends Component<
         }
     };
     private requestAddress = async (value: string) => {
-        const GEOCODER_ADDRESS = constants.GEOCODER_ADDRESS;
+        const ADDRESS_GEOCODING_URL = constants.ADDRESS_GEOCODING_URL;
         const SEARCH_RESULT_COUNT = constants.ADDRESS_SEARCH_RESULT_COUNT;
         const center = this.map.getCenter();
         const lat = center.lat;
         const lng = center.lng;
-        const requestUrl = `${GEOCODER_ADDRESS}?text=${value}&size=${SEARCH_RESULT_COUNT}&focus.point.lat=${lat}&focus.point.lon=${lng}`;
+        const requestUrl = `${ADDRESS_GEOCODING_URL}?text=${value}&size=${SEARCH_RESULT_COUNT}&focus.point.lat=${lat}&focus.point.lon=${lng}`;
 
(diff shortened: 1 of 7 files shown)


Divinare:
WIP: postNumber should be queried from digitransit API instead. Openstreetmap's postnumber is not that reliable

Divinare:
This can now be reviewed.

ahjyrkia:
Bug with the stop form:

1) Create new node called "first".
2) Create another new node called "second".
3) Refresh page (and url parameters).
4) Click the "first" node on map and then click the "second" node.
5) Close "second" node by pressing "X" on top right of sidepanel.
6) The map focuses on "first" node and opens it but it has values from the "second" node. 

This problem persists even if you try to open a different node and come back to the "first" node.

Divinare:
Stop form bug will be fixed here: https://github.com/HSLdevcom/jore-map-ui/issues/1034

ahjyrkia:
Works and reviewed. A bug found that is made in to its own issue

PR #1028, opened 2019-10-04
Fetching routePath after save, some refactoring

Divinare:
Changes:
* routePath & viaNames are fetched after save
* viaNameHash -> viaName array
* refactored setRoutePath() -> init()

How to test:
* Have backend in branch 874-kilpiVia with latest changes
* Test updating existing routePath
    * check that old viaNames with old routePathLinkIds are being destroyed from jore DB (remember, routePathLinks are re-created currently. This is why old viaNames need to be destroyed and new ones created with new routePathLink ids)
* Test creating new routePath

### DIFF
--- a/src/components/sidebar/routePathView/RoutePathView.tsx
+++ b/src/components/sidebar/routePathView/RoutePathView.tsx
@@ -121,12 +121,13 @@ class RoutePathView extends ViewFormBase<
                     lineId,
                     route
                 );
-                this.props.routePathStore!.setRoutePath(newRoutePath);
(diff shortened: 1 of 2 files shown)


ahjyrkia:
There seems to be a problem with updating routepath's links. How to recreate the bug:
* Find a link that has some viaName data. 
* Edit and save a different link within the same routePath.
* Look at the link that previously had viaName data. The data is no longer there.

Divinare:
Fixes done. Remember to also update 874-kilpiVia backend PR when testing.

PR #1030, opened 2019-10-04
WIP: Link length calculated automatically

### LINKED ISSUE #892: Implement measured length for links & routePath
PR: https://github.com/HSLdevcom/jore-map-ui/pull/1030

Link
* lnkmitpituus (measuredLength): Manual input, can be left empty
* lnkpituus (length): Is calculated automatically according to jr_piste locations of a link

RoutePath
* suupituus (length: Base value is calculated by summing up measured length of each link. If measured length is missing for a link, use lnkpituus instead.

Node
* Changes will be done in #1036 

---

old planning: if saving link, should routePath's using the link suupituus be updated? Also if user opens routePath, should length be calculated manually? (so if there is a new length available, user just clicks edit pen and can save without changing anything)

Divinare:
Closes #892 

* Read card description
    * also removed ability to see which length is going to be used e.g "Laske 1050m". That button might be removed completely in future tasks so didn't want to invest time so that is why I removed it


Divinare:
Seems to work

PR #1049, opened 2019-10-14
1022 node link opening fix

### LINKED ISSUE #1022: Move isEditingDisabled out from ViewFormBase into store

Divinare:
Closes #1022 

Changes:
* Removed react warnings about missing key prop (dashed lines caused warnings)
* node form always opens in is editing disabled
* link form always opens in is editing disabled
* link editing undo bug fixed

Before testing, you can find those bugs in dev environment

link editing undo bug:
1) open a link
2) click edit pen TWICE
3) make two geometry changes
4) undo twice, it should return to the original position. Bug is, 3 undo's are required instead

### DIFF
--- a/src/components/map/layers/edit/EditLinkLayer.tsx
+++ b/src/components/map/layers/edit/EditLinkLayer.tsx
@@ -161,11 +161,13 @@ class EditLinkLayer extends Component<IEditLinkLayerProps> {
         const link = this.props.linkStore!.link;
         return [
             <DashedLine
+                key={'startNodeDashedLine'}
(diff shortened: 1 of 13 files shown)


ahjyrkia:
Link editing bug:

1) Open link and edit geometry once
2) Undo 
3) "Tallenna muutokset" button is indicating that there is changes to be saved

ahjyrkia:
Suggestion: editing could be disabled after saving.

ahjyrkia:
Works. Review done.

Issue #1050, opened 2019-10-14
Edit link, show what is being saved


Divinare:
If editing link, save button opens a popup window that lists difference between old link and current link to be saved. User can then click buttons [Peruuta] [Tallenna]

LinkView.tsx
`onClick={this.save} -> onClick={this.savePrompt}`

```
private renderConfirmContent = () => {
     const foo = ... // list differences between this._oldLink & this.link in linkStore
     return (
            <div>...foo...</div>
    )
}

private savePrompt = () => {
        ConfirmStore.openConfirm(this.renderConfirmContent(), () => {
            this.save();
            foo();
        });
}
```

PR #1057, opened 2019-10-16
1050 show what is being saved

ahjyrkia:
closes: https://github.com/HSLdevcom/jore-map-ui/issues/1050

To be continued in: https://github.com/HSLdevcom/jore-map-ui/issues/1066

### DIFF
--- a/src/components/overlays/SavePrompt.tsx
+++ b/src/components/overlays/SavePrompt.tsx
@@ -13,29 +13,6 @@ interface ISavePromptProps {
 }
 
 const SavePrompt = (props: ISavePromptProps) => {
-    const renderChangesContent = () => {
-        const newLink = _.cloneDeep(props.newData);
-        const oldLink = _.cloneDeep(props.oldData);
-        for (const property in newLink) {
-            const linkAttribute = newLink[property];
-            const oldLinkAttribute = oldLink[property];
-            if (_.isEqual(linkAttribute, oldLinkAttribute)) {
-                delete oldLink[property];
-                delete oldLink[property];
-            }
-        }
-
-        return Object.keys(oldLink).map((property: string, index: number) => {
-            const propertyLabel = getLabel(props.type, property);
-            return (
-                <div key={index} className={classnames(s.formItem, s.savePromptRow)}>
-                    <div className={s.inputLabel}>{propertyLabel}</div>
-                    {callRenderByProperty(property, oldLink, newLink)}
-                </div>
-            );
-        });
-    };
-
     const callRenderByProperty = (property: string, oldLink: Object, newLink: Object) => {
         switch (property) {
             case 'geometry':
@@ -80,10 +57,29 @@ const SavePrompt = (props: ISavePromptProps) => {
         );
     };
 
+    const newLink = _.cloneDeep(props.newData);
+    const oldLink = _.cloneDeep(props.oldData);
+    for (const property in newLink) {
+        const linkAttribute = newLink[property];
+        const oldLinkAttribute = oldLink[property];
+        if (_.isEqual(linkAttribute, oldLinkAttribute)) {
+            delete oldLink[property];
+            delete oldLink[property];
+        }
+    }
+
     return (
         <div className={s.savePromptView}>
             <div className={s.topic}>Tallennettavat muutokset</div>
-            {renderChangesContent()}
+            {Object.keys(oldLink).map((property: string, index: number) => {
+                const propertyLabel = getLabel(props.type, property);
+                return (
+                    <div key={index} className={classnames(s.formItem, s.savePromptRow)}>
+                        <div className={s.inputLabel}>{propertyLabel}</div>
+                        {callRenderByProperty(property, oldLink, newLink)}
+                    </div>
+                );
+            })}
         </div>
     );
 };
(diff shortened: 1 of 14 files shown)


src/components/overlays/SavePrompt.tsx
```diff
@@ -0,0 +1,48 @@
+import classnames from 'classnames';
+import _ from 'lodash';
+import React from 'react';
+import { FiArrowRight } from 'react-icons/fi';
+import { getLabel } from '~/codeLists/labelCodes';
+import * as s from './savePrompt.scss'; // should use savePrompt.scss
+
+const SavePrompt = (props: any) => {
+    const changesContent = () => {
```
Divinare: Our convention has been "if some method returns JSX, it should start with `render` prefix"

src/components/overlays/SavePrompt.tsx
```diff
@@ -0,0 +1,48 @@
...
+                            <div className={s.newAttribute}>{current[property]}</div>
+                        </div>
+                    </div>
+                </div>
+            );
+        }
+        return <div className={s.savePromptView}>{divs}</div>;
+    };
+
+    return <div>{changesContent()}</div>;
```
Divinare: ```<div className={s.savePromptView}>{changesContent()}</div>;```

src/components/overlays/SavePrompt.tsx
```diff
@@ -0,0 +1,48 @@
...
+        for (const property in current) {
+            const linkAttribute = current[property];
+            const oldLinkAttribute = old[property];
+            if (_.isEqual(linkAttribute, oldLinkAttribute)) {
+                delete old[property];
+                delete old[property];
+            }
+        }
+
+        for (const property in old) {
```
Divinare: Do not use for loop and push to divs list, its ugly..

Instead use .map, like this:
```
return properties.map((property: string) => {
    return (
            <div className={classnames(s.formItem, s.savePromptRow)}>
             .....
            </div>
    )
}
```

Divinare: Also you can use a method? renderChangeRow?

```
return properties.map((property: string) => {
     const oldString = ...
     const newString = ...
     return this.renderChangeRow(oldString, newString);
}
```

src/components/overlays/SavePrompt.tsx
```diff
@@ -0,0 +1,48 @@
...
+            const propertyLabel = getLabel(props.type, property);
+            divs.push(
+                <div className={classnames(s.formItem, s.savePromptRow)}>
```
Divinare: Editing link geometry causes a bug " Each child in an array or iterator should have a unique "key" prop". This prob is because of you are rendering rows in a loop without specifying a key for a row. One way to fix this is to use: <div key={`row-${index}`}></div> (index comes from map: ```return properties.map((property: string, index: number) => {```

ahjyrkia:
Requested changes done.

src/components/sidebar/linkView/LinkView.tsx
```diff
@@ -381,7 +391,11 @@ class LinkView extends ViewFormBase<ILinkViewProps, ILinkViewState> {
                         </div>
                     </Button>
                 </div>
-                <Button type={ButtonType.SAVE} disabled={isSaveButtonDisabled} onClick={this.save}>
+                <Button
+                    type={ButtonType.SAVE}
+                    disabled={isSaveButtonDisabled}
+                    onClick={this.savePrompt}
```
Divinare: showSavePrompt?

src/components/overlays/SavePrompt.tsx
```diff
@@ -0,0 +1,66 @@
+import classnames from 'classnames';
+import _ from 'lodash';
+import React from 'react';
+import { FiArrowRight } from 'react-icons/fi';
+import { getLabel } from '~/codeLists/labelCodes';
+import * as s from './savePrompt.scss';
+
+const SavePrompt = (props: any) => {
```
Divinare: Always try to avoid using any. Especially here, now we lose the benefits of typescript

src/components/overlays/SavePrompt.tsx
```diff
@@ -0,0 +1,66 @@
...
+            <div className={s.flexInnerRow}>
+                <div>Linkin geometria muuttunut.</div>
+            </div>
+        );
+    };
+
+    return (
+        <div className={s.savePromptView}>
+            <div className={s.topic}>Tallennettavat muutokset</div>
+            {renderChangesContent()}
```
Divinare: Does renderChangesContent() have to be in a separate method?

src/components/sidebar/linkView/LinkView.tsx
```diff
@@ -172,6 +175,18 @@ class LinkView extends ViewFormBase<ILinkViewProps, ILinkViewState> {
         }
     };
 
+    private savePrompt = () => {
+        const confirmStore = this.props.confirmStore!;
```
Divinare: if !isNewLink call this.save() straightly? Now if creating a new link, it shows prompt which is unnecessary.

ahjyrkia:
Fixes done

Divinare:
LGTM

PR #872, opened 2019-05-09
825 Set network date picker, and its default to now

kronken:
Closes #825 

Updates style of date picker in network selector, and make its default value as now

PR #1074, opened 2019-10-18
872 node view and link view is editing disabled status fixes

Divinare:
Closes #872

Fixed bug where in nodeView create new node you used create create new node tool again, new node form opened without editing enabled. Same happened with linkView.


ahjyrkia:
Works and reviewed

Issue #1048, opened 2019-10-11
Use browser's local storage to keep track of which layers are visible


Divinare:
* Save node and link layers to local storage

PR: https://github.com/HSLdevcom/jore-map-ui/pull/1071

PR #1110, opened 2019-10-30
1070 changes to app architecture, map coordinates init changes, app rendering problem fix

### LINKED ISSUE #1070: Changes to app architecture, map coordinates init changes, app rendering problem fix
* This can reduce initial load time significantly, from 6-7 seconds to 4 seconds (tested locally)
* Also the whole app usage speed is increased significantly if when extra rendering is avoided

Divinare:
Closes #1070 

As the result, speed of using the whole app increased significantly. Setting leaflet map coordinates architecture changed.

* the whole app re-rendered always when navigating to a new entity. For example, open a new line / lineHeader from home page -> no more map flashing! (you can compare this to our dev site, see how ugly it was)
* map initial position is set to Helsinki ONLY if views that want to center map to some objects don't prevent it
   * added mapStore.setIsMapCenteringPrevented // used to say "render map in non interactive mode, wait until map gets centered"
   * e.g. nodeFetch will center map, set setIsMapCenteringPrevented as false which enables using map
* fixed multiple code issues / hacks, did some refactoring
* Removed layers lazy load done in #1048
* bonus: node gets centered to all the links that belong to the node

The only bad thing imo is that views now need to center map to the objects they fetch. A better way would be to do this in layers instead. However, I couldn't find a way to implement that. It would be possible with always setting map initial position to Helsinki but that is what we don't want from user's perspective.


Divinare:
Self reviewed & tested.

ahjyrkia:
Tested and reviewed post merge.

PR #1114, opened 2019-10-31
Update link length when isEditing is enabled

### LINKED ISSUE #1105: Link automatic length causing prompt when clicking back button
Steps;
1) Open a link where calculated length changes
2) Clicking back button says "You have unsaved changes"

--> calc length only when edit button is clicked

Divinare:
Closes #1105


Divinare:
Self testing: seems to work.

Issue #1099, opened 2019-10-28
Remove "sovitettu piste" from UI, save it the same as mitattu piste


Divinare:
FE PR: https://github.com/HSLdevcom/jore-map-ui/pull/1117
BE PR: https://gitlab.hsl.fi/jore/jore-map-backend/merge_requests/142

PR #1117, opened 2019-11-04
1099 remove coordinatesManual

ahjyrkia:
closes https://github.com/HSLdevcom/jore-map-ui/issues/1099

BE pr: https://gitlab.hsl.fi/jore/jore-map-backend/merge_requests/142


Divinare:
Using geojsonManual (manual location) if geojson (measured location) is missing. Sometimes that is the case.

Reason is:
select * from jr_solmu where solomx IS NULL
returns some rows

select * from jr_solmu where solomx IS NULL AND solx IS NULL
returns no results

Divinare:
This can be reviewed now.

ahjyrkia:
Review done

Issue #1100, opened 2019-10-28
Short id search with jokers


Divinare:
`10*` gives results of: `1000, 1001, 1002, ... 1099`
`*001` gives results of `0001, 0002, 0003, ... 9001`

Remember, we already have jokers in line search --> reuse that code

PR #1121, opened 2019-11-05
1100 shortId search with jokers

ahjyrkia:
closes https://github.com/HSLdevcom/jore-map-ui/issues/1100

Known "bug": Dropdown input value is not editable after a value is selected. 
It's a pending Issue in the dependacy's repository: https://github.com/JedWatson/react-select/issues/2296

Added a card:
https://github.com/HSLdevcom/jore-map-ui/issues/1124

### DIFF
--- a/src/components/controls/Dropdown.tsx
+++ b/src/components/controls/Dropdown.tsx
@@ -147,7 +147,9 @@ class Dropdown extends React.Component<IDropdownProps, IDropdownState> {
                                 filterOption={null}
                                 placeholder={'Valitse...'}
                                 styles={_getCustomStyles(this.props)}
-                                noOptionsMessage={() => null}
+                                noOptionsMessage={() =>
+                                    props.isNoOptionsMessageHidden ? null : 'Ei hakutuloksia'
+                                }
                                 hideSelectedOptions={Boolean(this.props.isSelectedOptionHidden)}
                                 className={this.props.isBackgroundGrey ? s.greyBackground : ''}
                             />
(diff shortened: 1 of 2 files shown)


src/components/controls/Dropdown.tsx
```diff
@@ -133,11 +144,10 @@ class Dropdown extends React.Component<IDropdownProps, IDropdownState> {
                                 options={displayedItems}
                                 isDisabled={props.disabled}
                                 isSearchable={true}
+                                filterOption={null}
                                 placeholder={'Valitse...'}
                                 styles={_getCustomStyles(this.props)}
-                                noOptionsMessage={() =>
-                                    props.isNoOptionsMessageHidden ? null : 'Ei hakutuloksia'
-                                }
+                                noOptionsMessage={() => null}
```
Divinare: Why?

ahjyrkia: Forgot to change it back after debugging. Though after testing I feel like it could be better like this.

The message serves no purpose. When there is no search results to be shown the dropdown options is closed and the input validation message "Lyhyttunnus on jo käytössä" can be seen which is basically the same thing as saying "Ei hakutuloksia". Also if the user types in an actual validation error (i.e bad format, too long) and the validation error is hidden under the "Ei hakutuloksia" message.

Divinare: How about other dropdowns? All dropdowns are search enabled by default.

Divinare: Also if this is really always defined as `noOptionsMessage={() => null}` then current props isNoOptionsMessageHidden serves no purpose and should be removed. I would revert this change though.

ahjyrkia: Reverted

PR #1144, opened 2019-11-14
1122 mass edit line header topics

### LINKED ISSUE #1122: Mass edit lineHeader topics part 2, lineHeaderForm under lineHeader table
FE PR: https://github.com/HSLdevcom/jore-map-ui/pull/1144
BE PR: https://gitlab.hsl.fi/jore/jore-map-backend/merge_requests/146

- massEdit lineHeaders where lineHeader can be edited or removed
- lineHeader form should be displayed under the table at lineView page (remove /lineHeader endpointPath
- updatedAt fields should only change from those lineHeaders that were really edited.

Divinare:
Closes #1122 

BE PR: https://gitlab.hsl.fi/jore/jore-map-backend/merge_requests/146

Done some new changes to line header table:
* refactored IMassEditLineHeader interface, code is much more simplified now
* can't remove all lineHeaders (there must be at least one lineHeader)
* highlighting active lineHeader (active means that there is no lineHeader with start time / end time being between current timestamp)
    * new validation: can't save if there is no active lineHeader
* close prompt fixes
* bug fixes


Divinare:
Can be reviewed.

Some big changes here:

* LineHeader form appears now under lineHeaderTable
* New lineHeaderSaveModel that backend takes
* Removed old create & update lineHeader endPoints, removed old lineHeaderView

ahjyrkia:
Bug found:
Have two lineheaders with no validation errors. Copy the first one as new lineHeader. Third lineHeader is duplicate of first one and shows validation errors because of the dates. Remove the original lineHeader of the two duplicates. Now there's validation errors even though there shouldn't be

image

image

src/components/sidebar/lineView/lineHeaderTableRows.scss
```diff
@@ -0,0 +1,58 @@
...
+        }
+    }
+
+    .removeLineHeaderButton:hover {
+        color: white;
+        background-color: $red;
+    }
+}
+
+.lineHeaderRowHighlight {
```
ahjyrkia: Could we have border-radius here and some padding on the left side of the div?

src/services/lineHeaderService.ts
```diff
@@ -46,18 +57,54 @@ class LineHeaderService {
...
+    public static massEditLineHeaders = async (
+        massEditLineHeaders: IMassEditLineHeader[],
+        oldLineHeaders: ILineHeader[],
+        lineId: string
+    ) => {
+        const added: ILineHeader[] = [];
+        const edited: ILineHeader[] = [];
+        const removed: ILineHeader[] = [];
+        const originals: ILineHeader[] = [];
+        massEditLineHeaders!.forEach((massEditLineHeader: IMassEditLineHeader) => {
```
ahjyrkia: Almost the same but maybe a bit more readable like this:
```
        massEditLineHeaders!.forEach((massEditLineHeader: IMassEditLineHeader) => {
            const currentLineHeader = massEditLineHeader.lineHeader;
            if (massEditLineHeader.isRemoved) {
                removed.push(currentLineHeader);
                return;
            }

            const originalLineHeader = currentLineHeader.originalStartDate
                ? oldLineHeaders.find(oldLineHeader =>
                      areDatesEqual(
                          oldLineHeader.originalStartDate!,
                          currentLineHeader.originalStartDate!
                      )
                  )
                : null;

            if (!originalLineHeader) {
                added.push(massEditLineHeader.lineHeader);
            } else {
                _.isEqual(originalLineHeader, currentLineHeader)
                    ? originals.push(currentLineHeader)
                    : edited.push(currentLineHeader);
            }
        });
```

src/util/dateHelpers.ts
```diff
@@ -0,0 +1,20 @@
+import Moment from 'moment';
+
+const getDateString = (date: Date): string => {
```
ahjyrkia: Maybe rename to toDateString?

src/util/dateHelpers.ts
```diff
@@ -0,0 +1,20 @@
...
+    return Moment(date).format('DD.MM.YYYY');
+};
+
+const getDateWithoutHours = (date: Date): Date => {
```
ahjyrkia: Rename to something like roundDateToMidnight?

src/services/lineHeaderService.ts
```diff
@@ -46,18 +57,54 @@ class LineHeaderService {
...
+            const currentLineHeader = massEditLineHeader.lineHeader;
+            if (massEditLineHeader.isRemoved) {
+                removed.push(currentLineHeader);
+            } else {
+                const originalLineHeader = currentLineHeader.originalStartDate
+                    ? oldLineHeaders.find(oldLineHeader =>
+                          areDatesEqual(
```
ahjyrkia: I think _.isEqual works here too

src/components/sidebar/lineView/lineHeaderTableRows.scss
Divinare: There is already 2px padding on both sides. Increasing padding isn't a good idea since the space is already very limited.

I don't like border-radius there

src/util/dateHelpers.ts
Divinare: toMidnightDate seems to go better with toDateString

src/services/lineHeaderService.ts
Divinare: _.isEqual takes into account that date objects are exactly the same. That "areDatesEqual" compares year, month and date. It doesn't take into account seconds / hours . Even though dates should be converted to "midnightDate" this might not always be the case. 

Maybe in areDatesEqual() method there should be comparison with year / month / date + _.isEqual and if those results are different, it should throw something to prevent errors... additional check would slow down code execution though

Divinare:
Review fixes done.

About bug: "Also could it be possible when clicking "Luo uusi linjan otsikko" the default start and end dates would be the ending date of the last item in the list."
   -> Yeah, good idea. Those start & end dates are now last item's end date + 1 by default
   -> Also line header creation with copy button use the same logic with default dates

ahjyrkia:
Changes ok

Issue #1129, opened 2019-11-08
Create new stop stopArea not working


Divinare:
Steps:
1) open new stop form
2) select stopArea
3) deselect stopArea
4) --> name is not reseted

Solution: make stopArea always as a required field. Add a validation text under pysäkkialue dropdown

PR #1161, opened 2019-11-19
1129 create new stop area not working

ahjyrkia:
closes https://github.com/HSLdevcom/jore-map-ui/issues/1129


Divinare:
Made stopArea (i) button hidden if stopArea is not selected.

Review ok now

PR #1163, opened 2019-11-19
1137 radius circle can be clicked through

ahjyrkia:
closes https://github.com/HSLdevcom/jore-map-ui/issues/1137


Divinare:
Seems to work

PR #1176, opened 2019-12-02
1108 map objects selection with popup

### LINKED ISSUE #1108: Make it easier to select nodes and links from map
Postgis PR: https://gitlab.hsl.fi/jore/jore-map-postgis/merge_requests/28
FE PR: https://github.com/HSLdevcom/jore-map-ui/pull/1176

* Show nodes and links on popup near click location
   * highlight links / nodes

How to implement
* How about having "sidebarOverlaytStore" with "openSidebarOverlay()" like in confirmStore. That would be an overlay to the sidebar.
* perhaps a better solution: show a popup with those objects that were near click. Hovering an option shows the option. Also the popup could have some little opacity (option as props) so that elements behind popup gets hovered as well.

Divinare:
Closes #1108 

Postgis PR: https://gitlab.hsl.fi/jore/jore-map-postgis/merge_requests/28

Popup to show links & nodes around click
* Implemented with POSTGIS func call (couldn't find a way to do this completely in frontend)
* Changed a bit how popups are rendered. Now you can pass data=any & type="popupType" to "openPopup" func that renders a popup component based on given popupType and passes data for that popup component. Popups can still be created with passing reactNode. Maybe there should be only one way to render a popup.

To fix hardcoded bufferSize: https://github.com/HSLdevcom/jore-map-ui/issues/1177

TO TEST:

pull postgis PR, run ./updatedb.sh (NOTIFY: this should now never get stuck! I fixed a delete query there that took like 15 - 45 min or sometimes was stuck forever)


ahjyrkia:
Would it be possible to show nodes and links based on what layers are open? E.g if link layer is open only show links in the new popup

Divinare:
Nodes are shown only if nodeLayer is open. Forgot to implement the same thing with links

src/components/map/layers/popups/SelectNetworkEntityPopup.tsx
```diff
@@ -0,0 +1,108 @@
...
+                    );
+                })}
+                {this.props.data.links.map((link: ILinkMapHighlight, index: number) => {
+                    return (
+                        <div
+                            key={index}
+                            className={s.row}
+                            onMouseOver={() => this.highlightLink(link, true)}
+                            onMouseOut={() => this.highlightLink(link, false)}
+                            onClick={() => this.redirectToLink(link, this.props.popupId)}
```
ahjyrkia: Should we have savePrompt triggered here as well?

ahjyrkia:
changes ok

Issue #1190, opened 2019-12-04
Saving stopArea saves names of stops belonging to stopArea


Divinare:
FE PR: https://github.com/HSLdevcom/jore-map-ui/pull/1196
BE PR: https://gitlab.hsl.fi/jore/jore-map-backend/merge_requests/157

How to implement
* FE: at save prompt, show text "Tallentaminen muuttaa kaikkien pysäkkialueeseen kuuluvien pysäkkien nimet: 100, 101, 102"
* BE needs to save stops, sync etc.
* FE: when user opens nodeView, always fetch newest data. If a stopCache is used, replace oldNode with a node that was just fetched.

Also make sure this stopArea bug is fixed:

Steps:

1) Select node and create new stopArea
2) Save
3) Select another node, toggle edit and save
4) savePrompt has same changes again

1) Select node and create new stopArea
2) Save
3) Select another node, toggle edit and save
4) savePrompt has same changes again

PR #1191, opened 2019-12-05
1175 show stops on map that belong to stop area

### LINKED ISSUE #1175: Show stops that belong to a stopArea on map (when stopAreaView is opened)
FE PR: https://github.com/HSLdevcom/jore-map-ui/issues/1175

Clicking a stop in stopList centers map into that stop on map. Also highlight it when hovered (?). Before implementing highlight, make sure #1108 is merged into this task (#1108 might implement 
a highlight store that can be utilized here)

Divinare:
Closes #1175 

* Had to rewrite `NodeMarker` so that it is more flexible
   * You can call it without node object (look at StopAreaLayer, it can now use `NodeMarker`!)
   * Renders only one LeafletMarker, no more ugly double marker rendering
* Had to rewrite `CoordinateControl` as a react component so that it doesn't require mapStore to have coordinates (in previous implementation, leaflet's onAdd was only called once and it required coordinates to be set. Even if there was a way, this rewriting made the component better customizable)
* IStopItem moved to models/stop (it is a stop variant and a model, used at multiple places)
* Centering map straightly into stopArea's stops. For example, you can go to http://localhost:3000/stopArea/110604
   * See how CoordinateControl component has no x y at first, it now supports empty map coordinates
* Using new typescript 3.7+ syntax here: `const y = this.state?.y || '';`
    * This means the same as:
```
const y = this.state && this.state.y ? this.state.y || '';`
```

Testing: test all variants of
* nodeView, edit & drag nodes
* coordinateControl
* stopAreaView

etc.


ahjyrkia:
Bugs:

1) Select node and create new stopArea
2) Save
3) Select another node end press edit toggle
4) savePrompt has same changes again

ahjyrkia:
Review done

Divinare:
That bug fill be fixed here: https://github.com/HSLdevcom/jore-map-ui/issues/1190

PR #1194, opened 2019-12-11
1140 let user input date freely

### LINKED ISSUE #1140: Let user input date freely by typing
How to implement?

at InputContainer.tsx:

```
        // if (Moment(value, 'DD.MM.YYYY', true).isValid()) {
        //     onChange(value);
        // }
        onChange(value);
```

then, if input container gets a value that is not a date, display normal input (without calendar)? Or maybe better, dont give selected value to calendar component

Divinare:
Closes #1140 

input can now be typed freely.
* the caveat: react-datepicker uses autocomplete which cant be currently turned off (react-datepicker doesnt support turning it off) Will be fixed at new task when that autocomplete param works: fix autocomplete
* DatePicker is now in its own component
    * has its own selected date, that is passed to parent ONLY if date is a valid date
         * invalid date field will become red. User can still save but the date he is using is old date!
    * supports empty date (that will now become red if left empty) via props: isEmptyValueAllowed
        * nodeForm's "mittauspvm" and maplayerControl's date uses this


ahjyrkia:
Post merge reviewed

PR #1197, opened 2019-12-13
1183 node view stop using view form base

Divinare:
Notified that there are common things in all Views:
* isEditingDisabled change -> have to call validateAllProperties()
* object to validate (eg. ILink)
    * object properties have to have listeners and they have to be validated when they change
         * when a object to validate changes, existing listeners need to be removed
    * need access to validation result
* validation model (eg. ILinkValidationModel)
* invalidPropertiesMap
    * some properties may require custom validation functions where result depends on values of other objects. Some properties may require other properties to validate when they are validated

--> implemented a generic ValidationStore to handle these requirements.

When user puts a letter into inputContainer, currenly flow goes like this:
InputContainer -> StopView -> StopStore (update stop property) -> ValidationStore -> StopStore (updated invalidPropertiesMap)-> Stop View -> InputContainer (props.value is updated)

How to test:
* test that nodeView works (stop & node form)
* test that lineView works
* test that switching fast between nodeViews doesn't cause any warnings / thrown errors

### DIFF
--- a/src/components/sidebar/lineView/LineInfoTab.tsx
+++ b/src/components/sidebar/lineView/LineInfoTab.tsx
@@ -24,7 +24,6 @@ interface ILineInfoTabProps {
     isEditingDisabled: boolean;
     isNewLine: boolean;
     isLineSaveButtonDisabled: boolean;
-    invalidPropertiesMap: object;
     saveLine: () => void;
 }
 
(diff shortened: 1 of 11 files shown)


ahjyrkia:
Post merge reviewed

PR #1199, opened 2019-12-16
1169 stop using ViewFormBase in linkView, routePathView, stopAreaView

### LINKED ISSUE #1169: RoutePathView, RouteView, LinkView, stopAreaView: stop using ViewFormBase
Blocked by #1183 

Use validationStore implemented in #1183

This solves bugs such as:

Steps:
1) Open 3 different nodes for example
2) spam browsers back button

--> getting error TypeError: this.isEditingDisabledListener is not a function

Divinare:
Closes #1169 

* IViaName properties are now part of IRoutePathLink properties (they are extracted out before sending to backend - I know, they could be send to backend in the same object but it isn't worth refactoring)
    * code is a lot cleaner this way, IViaName validation is part of IRoutePathLink validations
* Had to change sidebar & map rendering order. Reason: map needs to be rendered before <Sidebar /> so that listeners get initialized before Views set map coordinates.
* Now if validating two properties at once, validation error will be only shown to the first field (prevents duplicate errors)

### DIFF
--- a/src/components/App.tsx
+++ b/src/components/App.tsx
@@ -60,6 +60,7 @@ class App extends React.Component<IAppProps, IAppState> {
             <>
                 <NavigationBar />
                 <div className={s.appContent}>
+                    {/* Map needs to be rendered before <Sidebar /> so that listeners get initialized before Views set map coordinates. */}
                     <Map>
                         <ErrorBar />
                     </Map>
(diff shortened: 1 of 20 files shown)


ahjyrkia:
Clicking open few nodes and then spamming browser back button produces warning: 

```Warning: Can't perform a React state update on an unmounted component. This is a no-op, but it indicates a memory leak in your application. To fix, cancel all subscriptions and asynchronous tasks in the componentWillUnmount method.
    in NodeView (created by inject-NodeView-with-alertStore-nodeStore-mapStore-errorStore-confirmStore)
    in inject-NodeView-with-alertStore-nodeStore-mapStore-errorStore-confirmStore
    in Unknown (created by Route)```

Issue #1201, opened 2019-12-17
Fix codeListStore init to improve app speed


Divinare:
Now each time getDropdownItemList() is called, 1300 dropdown items are iterated through which causes huge input lag

Rewrite codeListStore init to iterate list only once, store items into map

ahjyrkia:
Reviewed

PR #1202, opened 2019-12-17
1102 stop creation outside of local area

### LINKED ISSUE #1102: Stop creation outside of localArea geojson
FE PR: https://github.com/HSLdevcom/jore-map-ui/pull/1202
BE PR: https://gitlab.hsl.fi/jore/jore-map-backend/merge_requests/158

* How is this done in Kouvola?

PLAN:

* StopForm: make it possible to unselect transitType

When opening node form -> query.isValid(point)
- if found, returns nodeId -> UI: show as non editable text field
- else returns null -> UI: show error: "valitulle pisteelle ei löydy aluetietoja, täytä solmun 5 ensimmäistä numeroa käsin."
   - user can edit first 5 letters, after 5 letters exists -> query.getNodeIds(beginningOfNodeId)
       -> returns list [00, 01, 02, ...] OR list of 7-letter ids
      -> if empty array [], UI: show validation error "solmun tunnusta ei löytynyt"

Divinare:
Closes #1102 

BE PR: https://gitlab.hsl.fi/jore/jore-map-backend/merge_requests/158

Flow described in card

Added idSuffix to INode. This way code is a lot cleaner (no explicit code for handling idSuffix is needed). Just have to make sure that when node is created, final nodeId saved as id + suffix

How to test:

* Try to create a new node in Helsinki
    * try to switch transitType -> it should re-render nodeView and generates a new nodeId
* Try to create a new node in Kouvola
    * notification should be shown that node id automatic generation has failed
    * input 5 letters, select ending from dropdown

There is some nodeId input lag. Will be fixed here: https://github.com/HSLdevcom/jore-map-ui/issues/1201

### DIFF
--- a/src/codeLists/propertyCodeLists.ts
+++ b/src/codeLists/propertyCodeLists.ts
@@ -4,6 +4,7 @@ type NodeKeys = keyof INode;
 type INodePropertyCodeList = { [key in NodeKeys]: string };
 const nodePropertyCodeList: INodePropertyCodeList = {
     id: '',
+    idSuffix: '',
     shortIdLetter: 'LYHYTTUNNUS (2 kirj.)',
     shortIdString: 'LYHYTTUNNUS (4 num)',
     type: 'TYYPPI',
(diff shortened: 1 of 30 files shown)


ahjyrkia:
image
This input is not restored when making coming back from editing/creating stopArea

PR #1210, opened 2019-12-20
CodeListItems are now shorted only at init

### LINKED ISSUE #1201: Fix codeListStore init to improve app speed

Divinare:
Closes #1201 

How to test:
* inputing numbers should be now much faster


ahjyrkia:
Reviewed

PR #1233, opened 2020-01-10
Navigator.goTo shows unsaved changes prompt

### LINKED ISSUE #1228: Refactor showing change view prompt
FE PR: https://github.com/HSLdevcom/jore-map-ui/pull/1233

Currently e.g. homePage button doens't show prompt, need a more generic solution:

Create a new navigationStore

Add to: navigator.goTo()
check if (navigationStore.shouldShowPrompt) { }
* show prompt:

if user accepts: this.store.history.push(url);
if user denies: return;

Each view listens to isDirty, if true -> navigationStore.setShouldShowPrompt(true);

Why refactor? This way each link automatically uses prompt.

Remember to remove showing prompt from SelectNetworkEntityTool, remove ToolbarStore!.shouldShowEntityOpenPrompt

Divinare:
closes #1228 

* Navigator.goTo params refactored as object
* Now every time navigator.goTo is used, prompt will be shown if current store has isDirty = true except when shouldSkipUnsavedPrompt: true is given
* nodeStore's ClearNodeCache method fixed so that it only clears that cache for the currently opened node (before it cleared the whole cache)

Times when unsaved changes prompt is not shown:
* Moving between nodeView & stopAreaView
* Clicking log out button
* Saving & redirecting to new entity

Test that prompts work in every view, every button that changes view. As an example, routePathView's routePathLinks now show prompts!

### DIFF
--- a/src/stores/navigationStore.ts
+++ b/src/stores/navigationStore.ts
@@ -1,0 +1,23 @@
+import { action, computed, observable } from 'mobx';
+
+export class NavigationStore {
+    @observable private _shouldShowPrompt: boolean;
+
+    constructor() {
+        this._shouldShowPrompt = false;
+    }
+
+    @computed
+    get shouldShowUnsavedChangesPrompt(): boolean {
+        return this._shouldShowPrompt;
+    }
+
+    @action
+    public setShouldShowUnsavedChangesPrompt = (shouldShowPrompt: boolean) => {
+        this._shouldShowPrompt = shouldShowPrompt;
+    };
+}
+
+const observableNavigationStore = new NavigationStore();
+
+export default observableNavigationStore;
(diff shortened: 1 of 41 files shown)


ahjyrkia:
X close button doesn't work when warning prompt is shown:

1) Open node and change something
2) Warning prompt is shown
3) Click "kyllä" 
4) Nothing happens

src/stores/navigationStore.ts
```diff
@@ -0,0 +1,23 @@
+import { action, computed, observable } from 'mobx';
+
+export class NavigationStore {
+    @observable private _shouldShowPrompt: boolean;
```
ahjyrkia: Maybe rename to just _showPrompt

Could also rename all the "shoulds" below

Divinare: Don't want to start booleans with a verb, otherwise you can't know if its a method or boolean

PR #1241, opened 2020-01-13
Fetch first and last routePath's stop names

### LINKED ISSUE #1207: Change routePath topic at routeList page
Blocked by #1179 
FE PR: https://github.com/HSLdevcom/jore-map-ui/pull/1241

NOTE: Earlier plan was to also change route topic at SearchView but is not that easilly doable (to prevent lag, would have to add first stop & last stop names to route table). Also it would be a bit strange when it would be first routePath's first stop & last stop.

Query flow to get first & last stop names of a route:
routePath -> firstRoutePathLink -> node by lnkalkusolmu -> stop by node id

How to implement:
Perhaps add to graphqlQueries:
getFirstRoutePathLinkQuery()
getLastRoutePathLinkQuery()

NOTE:
Here you can expect that routePath always ends to a stop and always starts with a stop

Divinare:
Closes #1207 

* Fetched routePathLinks need to be sorted because sometimes they are not ordered by orderNumber (e.g. tram http://localhost:3000/routes/?routes[0]=1007M)
* Finding the first / last stop that is not a crossroad. Yes, I could take "relpysakki" (is stop in use) into account but its not worth the time implementing. Also maybe it is a feature, not a bug

### DIFF
--- a/src/components/shared/loader/Loader.tsx
+++ b/src/components/shared/loader/Loader.tsx
@@ -8,6 +8,7 @@ type loaderSize = 'tiny' | 'small' | 'medium';
 interface ILoaderProps {
     size?: loaderSize;
     hasNoMargin?: boolean;
+    containerClassName?: string;
 }
 
 const Loader = observer((props: ILoaderProps) => (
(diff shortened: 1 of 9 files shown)


ahjyrkia:
Introduces a bit of lag (try using the map during fetching) when there's a lot of queries to do like here http://localhost:3000/routes/?routes[0]=1506&routes[1]=1010 but I guess it's quite a niche problem since usually there is not that many routePaths and anyways they are only fetched once. Other than that everything seems to be ok. 

If there's no quick and easy solution to the lagging this can be merged.

Divinare:
Optimization fixes done. Now it lazy loads only stopNames for those routePaths that are currently displayed. Also the map is centered straightly into routePaths.

Should see a huge difference.

ahjyrkia:
Review doned

PR #1243, opened 2020-01-14
Fix save prompt bugs

Divinare:
Had to disable searchBar if isEditing at routesPage. Reason: selecting a new route caused routes page to re-render (routeStore gets cleared due to routeList unmount call). Surely there could be a better fix but this is what I came up with.

Had to simplify the Navigator / sidebarHeader idea:

- navigator
    - close (might show prompt)
    - go back (might show prompt)

- sidebarHeader
    - if isEditPromptHidden = true, calls custom onEditButtonClick method
    - if isEditPromptHidden = true, might show prompt, then calls custom onEditButtonClick method

    - default close, calls navigator.close (can pass custom closeButtonMessage to navigator.close)
    - default back, calls navigator.goBack (can pass custom backButtonMessage to navigator.goBack)

    - calls custom onCloseButtonClick if method is given (might show prompt if defined at onCloseClick callback)
    - calls custom onBackButtonClick if method is given (might show prompt if defined at onBackButtonClick callback)

Test carefully. There are many combinations to test.

### DIFF
--- a/src/components/shared/searchView/SearchInput.tsx
+++ b/src/components/shared/searchView/SearchInput.tsx
@@ -1,3 +1,4 @@
+import classnames from 'classnames';
 import { inject, observer } from 'mobx-react';
 import React from 'react';
 import { SearchResultStore } from '~/stores/searchResultStore';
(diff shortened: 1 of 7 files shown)


ahjyrkia:
Reviewed. Could not find any bugs

PR #1269, opened 2020-01-20
Reduced decimals in geometries, fixed node savePrompt diff

### LINKED ISSUE #1128: Switch node type: store stop information in some distinct memory
FE PR: https://github.com/HSLdevcom/jore-map-ui/pull/1269
UPDATE: node location changes a bit because of projectedCoordinate is in different projection. Fix rounding decimals to 6, remove measurementType & coordinatesProjection from savePrompt if nodeType is crossroad / municipality border
* switch node -> stop causes issue: shows sovitettu & projisoitu piste changes when it shouldnt
   - one way: save node.stop to some distinct memory instead of keeping it at node.stop
* still switching from nodeType 'P' -> 'X', 'X' -> 'P' should show old stop data

Relates to: https://github.com/HSLdevcom/jore-map-ui/issues/1069

Divinare:
Closes #1128 

Reduced decimals to 6 because backend's columns have max 6 decimals: numeric(8,6)

### DIFF
--- a/src/constants/constants.ts
+++ b/src/constants/constants.ts
@@ -6,13 +6,15 @@ if (process.env.NODE_ENV === 'development') {
 } else {
     APP_URL = `https://${process.env.ENVIRONMENT}.${process.env.DOMAIN_NAME}`;
 }
-const environment: Environment = process.env.ENVIRONMENT ? process.env.ENVIRONMENT as Environment : Environment.LOCALHOST;
+const environment: Environment = process.env.ENVIRONMENT
+    ? (process.env.ENVIRONMENT as Environment)
+    : Environment.LOCALHOST;
 
 const commonConstants = {
     ENVIRONMENT: environment,
     BUILD_DATE: process.env.BUILD_DATE,
     AFTER_LOGIN_URL: `${APP_URL}/afterLogin`,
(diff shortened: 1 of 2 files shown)


src/constants/constants.ts
```diff
@@ -6,13 +6,15 @@ if (process.env.NODE_ENV === 'development') {
...
+const environment: Environment = process.env.ENVIRONMENT
+    ? (process.env.ENVIRONMENT as Environment)
+    : Environment.LOCALHOST;
 
 const commonConstants = {
     ENVIRONMENT: environment,
     BUILD_DATE: process.env.BUILD_DATE,
     AFTER_LOGIN_URL: `${APP_URL}/afterLogin`,
-    DECIMALS_IN_GEOMETRIES: 8,
+    DECIMALS_IN_GEOMETRIES: 6, // Max value 6 currenlty at joredb columns: numeric(8,6)
```
ahjyrkia: If you have 6 decimals in geometries and then you keep mashing a number that is 5 or larger the value seems to be changing. This probably has nothing to do with the amount of decimals allowed but some bug with the rounding of the number.

Divinare: Will be fixed here: https://github.com/HSLdevcom/jore-map-ui/issues/1322

PR #1285, opened 2020-01-27
1262 refactor util folder

### LINKED ISSUE #1262: Refactor util folder
* rename util -> utils
* don't use export new when not needed (e.g. ApiClient). If component has not state, it doesnt have to use new obj()
* rename ApiClient -> HttpUtils (check other naming as well)
* rename those utils that have state as helpers, e.g. ApolloClient -> ApolloClientHelper
* move helpers into /helpers
* remove those helpers/utils that are only used once (if they are too specific) to reduce noise in utils/helpers folder
* update contribution.md

Divinare:
Closes #1262


ahjyrkia:
Reviewed

PR #1732, opened 2020-10-16
Stage uses now hslid prod url

Divinare:


Divinare:
Actually this change can't be done. We dont want to use prod hsl id in stage (existing users with dev hsl id accounts wouldnt be able to login)

